PR #11, opened 2016-04-14
custom instrumentation docs did not work with v1.1.2. updated them

spnkr:
e.g. `Appsignal::Transaction.current.complete!` is not a valid method.


thijsc:
Thank you!

PR #13, opened 2016-04-15
Advise against tagging with personal information

jornvandebeek:
There are already public Github projects to be found that use the tag_request functionality to add personally identifiable information to exceptions. It is my understanding that Appsignal usually advises against passing through personally identifyable or otherwise sensitive information to the service.

The documentation thus should not use an example that encourages tagging an email address onto an exception.


wesoudshoorn:
@jornvandebeek sorry for the long wait and thanks for your PR. We have changed some wording in #18 because we strongly advise people not to send us personal data, but in the end it is up to our customers to use our features in a way that they see fit.

PR #16, opened 2016-05-09
Fixes custom job instrumentation example

mezis:
- `Appsignal::Transaction.create` takes 3 arguments.
- `Appsignal::Transaction` responds to `complete`, not `complete!`.


wesoudshoorn:
Thanks for fixing this @mezis !

PR #20, opened 2016-06-15
Suggest reporting javascript errors even when error is null

timraasveld:
Before this change, `window.onerror` would silently fail and report nothing to Appsignal if `error` was null.


matsimitsu:
👍  Awesome, thanks!

PR #23, opened 2016-06-21
Suggest using safe navigation for error.stack

timraasveld:
Apperantly, `error.stack` is not standard and will be unavailable in some cases, which was causing us an exception-in-an-exception

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Error/stack


thijsc:
Thanks for the contribution!

PR #27, opened 2016-07-26
Added extra information on how to jRuby

lucasmartins:
I've tested this tips at my production environment.


thijsc:
We've tweaked our SSL setup and you can now use JRuby without these settings, so I'm closing this one. Thanks for the contribution though, would have definitively merged it if the situation hadn't changed!

PR #28, opened 2016-08-01
Add hostname config documentation

tombruijn:
Adds documentation for the new `APPSIGNAL_HOSTNAME` env and config variable.

### DIFF
--- a/source/gem-settings/configuration.html.md
+++ b/source/gem-settings/configuration.html.md
@@ -23,7 +23,8 @@ This app's display name. If you use  Rails the gem will auto-detect the name and
 ## `APPSIGNAL_APP_ENV`
 This overrides the app's environment. Mostly used on Heroku where all apps run the `production` environment by default. This setting allows an override to set it to `staging` for example.
 
-## `APPSIGNAL_HOSTNAME`
+## `APPSIGNAL_HOSTNAME` / `:hostname` (since version 1.3)
+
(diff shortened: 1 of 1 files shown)


thijsc:
You should mention that this is available in 1.3 and up, this will be the next release.

source/gem-settings/configuration.html.md
```diff
@@ -23,6 +23,9 @@ This app's display name. If you use  Rails the gem will auto-detect the name and
 ## `APPSIGNAL_APP_ENV`
 This overrides the app's environment. Mostly used on Heroku where all apps run the `production` environment by default. This setting allows an override to set it to `staging` for example.
 
+## `APPSIGNAL_HOSTNAME`
```
matsimitsu: can you add the yaml config name? (`:hostname`)

tombruijn:
@thijsc Should I mention that its next release in the documentation page? That would mean we'd have to update the page to remove that line once it's released. I prefer to only mention the version number.

And in what way should I mark it that it's available since version 1.3. It would be nice if it stood out from the rest. I wanted to use the notice, but it's a bit big for this.

image

matsimitsu:
maybe add it behind the config var? (since 1.3) ?

tombruijn:
I'm okay with that

matsimitsu:
👍

PR #29, opened 2016-08-10
No more manual middleware inclusion for Sinatra

tombruijn:
From: https://github.com/appsignal/appsignal-ruby-private/pull/297

Is this enough or do we also want to keep the pre-1.3.0 method in the documentation for older installs?


matsimitsu:
👍

PR #30, opened 2016-08-12
Document AppSignal powered parameter filtering

tombruijn:
Documents AppSignal powered parameter filtering

Also updated some anchor links, but I see we sporadically use them. Something to look into, because having direct links/anchors to everything helps with sharing information.

## Related issues

- https://github.com/appsignal/appsignal-ruby-private/pull/304


matsimitsu:
👍 after adding that it's 1.3 and up :)

PR #33, opened 2016-08-15
Document send_params config

tombruijn:
- Document the config option
- Document the addition of the ENV variable since gem 1.3


thijsc:
👍

PR #34, opened 2016-08-16
Document AppSignal's own instrumentation

tombruijn:
Document the addition of `Appsignal.instrument` and how people can use this to instrument their own code. It's an rewrite of the "Custom instrumentation" page, rewriting it to put `ActiveSupport::Notifications` as an alternative instead of the main way to instrument code.

Also adding more documentation on what event names are and do.

`Appsignal.instrument` was added in: https://github.com/appsignal/appsignal-ruby/commit/0143ac7b27c213266e3cc84f2038bb705b1f1244

### DIFF
--- a/source/tweaks-in-your-code/custom-instrumentation.html.md
+++ b/source/tweaks-in-your-code/custom-instrumentation.html.md
@@ -126,45 +126,6 @@ SELECT * FROM users WHERE email = 'hector@appsignal.com' AND password = 'iamabot
 SELECT * FROM users WHERE email = ? AND password = ?
 ```
 
-## <a href="#activesupport_notifications" name="activesupport_notifications">ActiveSupport::Notifications</a>
-
-If you're using an older gem version than version 1.3 and you're using Rails
-(more specifically ActiveSupport) you can use the built-in
-ActiveSupport::Notifications to instrument your code instead.
-
-The method for instrumenting your code using `ActiveSupport::Notifications`
-is very similar to how AppSignal does it. Using the article fetcher example
-again you can see the differences are quite small.
-
-```ruby
-class ArticleFetcher
-  def self.fetch(category)
-    ActiveSupport::Notifications.instrument('fetch.article_fetcher') do
-      # Download and process the articles
-    end
-  end
-end
-
-ArticleFetcher.fetch('Latest news')
-```
-
-It works for nested instrumentation calls as well.
-
-```ruby
-ActiveSupport::Notifications.instrument('fetch.article_fetcher') do
-  10.times do
-    ActiveSupport::Notifications.instrument('fetch_single_article.article_fetcher') do
-      # Fetch single article
-    end
-  end
-end
-```
-
-`ActiveSupport::Notifications` is highly flexible, you can instrument your code
-any way you like. More information about `ActiveSupport::Notifications` can be
-found in the
-[Rails API docs](http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html).
-
 ## <a href="#event_naming" name="event_naming">Event name construction</a>
 
 Event names are used for many things in the inner workings of AppSignal.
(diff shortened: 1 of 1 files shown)


source/tweaks-in-your-code/custom-instrumentation.html.md
```diff
@@ -29,20 +29,101 @@ end
...
+```ruby
+sql = 'SELECT * FROM posts ORDER BY created_at DESC LIMIT 1'
+Appsignal.instrument('fetch.custom_database', 'Fetch latest post', sql) do
+  # ...
+end
+```
+
+### `body_format`
+
+TODO: ???
```
tombruijn: need some help with this part 😓

source/tweaks-in-your-code/custom-instrumentation.html.md
```diff
@@ -13,14 +13,14 @@ That's already very useful, but wouldn't it be great if we could see
 measurements of specific pieces of code you suspect might influence your
 performance? Well, you can!
 
-When you instrument your code using `ActiveSupport::Notifications`
-AppSignal will pick up on those instruments. For example, you have to work
-with an external API that fetches articles for your homepage:
+When you instrument your code with AppSignal we'll pick up on those
+instruments. For example, you have to work with an external API that fetches
+articles for your homepage:
```
thijsc: This sentence is a bit unclear now. You instrument your code using Appsignal now, we're not picking up AS events.

source/tweaks-in-your-code/custom-instrumentation.html.md
```diff
@@ -29,20 +29,101 @@ end
...
+### `body`
+
+More details such as a database query that was used by the event.
+
+```ruby
+sql = 'SELECT * FROM posts ORDER BY created_at DESC LIMIT 1'
+Appsignal.instrument('fetch.custom_database', 'Fetch latest post', sql) do
+  # ...
+end
+```
```
thijsc: We need to put a notice here that all user defined content should be stripped from the body string before you send it to us.

source/tweaks-in-your-code/custom-instrumentation.html.md
thijsc: If you use `Appsignal::EventFormatter::DEFAULT` here the body is passed through unchanged. If you use `Appsignal::EventFormatter:: SQL_BODY_FORMAT` the body will be sanitized using our SQL lexer.

source/tweaks-in-your-code/custom-instrumentation.html.md
tombruijn: Oh ok, so that's changed as well

source/tweaks-in-your-code/custom-instrumentation.html.md
tombruijn: The exception is when you have an SQL string and can use `body_format = Appsignal::EventFormatter::SQL_BODY_FORMAT`, right?

thijsc: Yes, see comment just below this :-).

source/tweaks-in-your-code/custom-instrumentation.html.md
thijsc: Well you can still use the AS way, that's just not what we recommend anymore.

tombruijn: > When you add custom instrumentation to your code you'll be able to receive even
> more insights into your application. For example, you have to work with an
> external API that fetches articles for your homepage:

better?

source/tweaks-in-your-code/custom-instrumentation.html.md
```diff
@@ -29,20 +29,101 @@ end
...
+
+### `name`
+
+The name of the event that will appear in the event tree in AppSignal.
+[Read more on key naming](#event_naming).
+
+### `title`
+
+A more descriptive title of an event, such as `"Load user
+hector@appsignal.com"`. This value can be unique per event.
```
matsimitsu: Ideally this is not unique per event e.g. `Load user` or `Find user by Email`

tombruijn:
updated :)

matsimitsu:
👍

PR #35, opened 2016-08-18
Document Capistrano in supported frameworks

tombruijn:
- Documents that we support Capistrano
- Documents https://github.com/appsignal/appsignal-ruby/pull/119

Not sure if we want to document `appsignal_config`, but that's the only other thing I found our capistrano integrations listens to.


matsimitsu:
👍

PR #51, opened 2016-10-30
Add heading anchors and links

tombruijn:
This adds anchors back to every heading and also makes the headings
links. :with_toc_data from Redcarpet doesn't work because we make the
headings links, but also because we use a custom renderer and the way
Middleman loads custom renders, the config is lost for custom renderers.


tombruijn:
This works to fix the headings anchors, but also makes them easily clickable and shareable.

@wesoudshoorn from a design/ui perspective, would this be good enough or should we maybe include the links as pseudo elements that appear on hover like GitHub does for headings in READMEs?

appsignalappsignal-docs 2016-10-30 11-15-43 1

tombruijn:
I'm merging this because I need to link to anchors now.

@wesoudshoorn let me know if you want to change the design of this.

PR #52, opened 2016-10-31
Add terminology page to explain the words we use

tombruijn:
## To explain

- [x] Agent (Ruby & Elixir) & system agent (Rust)
- [x] Alerting / Notifications
- [x] API
- [x] API key
- [x] AppSignal vs Appsignal vs appsignal
- [x] Application (App)
- [x] Blog
- [x] Changelog
- [x] Configuration
- [x] Events
- [x] Environments
- [x] Errors
- [x] Instrumentation
- [x] Instrumentation events
- [x] Integration
- [x] Markers
- [x] Metadata
- [x] Metric
- [x] Organizations
- [x] Owners
- [x] Performance issue
- [x] Push API
- [x] Push API key
- [x] Ruby Magic
- [x] Sample
- [x] Stroopwaffles
- [x] Tags
- [x] Teams
- [x] Transaction
- [x] User account

Let me know if I missed anything so far

### DIFF
--- a/source/appsignal/terminology.html.md
+++ b/source/appsignal/terminology.html.md
@@ -1,0 +1,460 @@
+---
+title: "Terminology"
+---
+
+At AppSignal we use a lot of technical terms. Read on to learn more about the
+language of AppSignal.
+
+- [Agents](#agents)
+  - [System agent](#system-agent)
+- [Alerting](#alerting)
+- [API](#api)
+- [API key](#api-key)
+- [Applications](#applications)
+- [AppSignal vs Appsignal vs appsignal](#appsignal-vs-appsignal-vs-appsignal)
+- [Blog](#blog)
+- [Changelog](#changelog)
+- [Configuration](#configuration)
+- [Environments](#environments)
+- [Errors](#errors)
+- [Events](#events)
+- [Exceptions](#errors)
+- [Instrumentation](#instrumentation)
+- [Instrumentation events](#instrumentation-events)
+- [Integrations](#integrations)
+  - [Agent integrations](#agent-integrations)
+  - [Integrations with services](#service-integrations)
+- [Markers](#markers)
+- [Metadata](#metadata)
+- [Metrics](#metrics)
+- [Organizations](#organizations)
+- [Owners](#owners)
+- [Performance issues](#performance-issues)
+- [Push API](#push-api)
+- [Push API key](#push-api-key)
+- [Ruby Magic](#ruby-magic)
+- [Samples](#samples)
+- [Stroopwafels](#stroopwaffles)
+- [Stroopwaffles](#stroopwaffles)
+- [Teams](#teams)
+- [Transactions](#transactions)
+- [User account](#user-account)
+
+## Agents
+
+AppSignal uses agents to monitor applications and hosts. These agents can be
+language specific, like our [Ruby](/ruby/index.html) and
+[Elixir](/elixir/index.html) agents. These agents include hooks into frameworks
+and libraries to instrument code blocks such as database calls, file system
+calls and view rendering.
+
+The instrumentation data collected from agents is sent to the AppSignal
+servers. The [transaction](#transactions) data is processed and used to detect
+events worth [alerting](#alerting) users about.
+
+Every agent is specialized in instrumentation of its subject language. Most
+AppSignal agents also includes a "[system agent](#system-agent)" which the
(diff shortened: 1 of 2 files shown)


wesoudshoorn:
@tombruijn good initiative! You might want to take some extra effort to explain the difference between the API key and the Push API key, because these both show up in the interface if I am correct and might be confused sometimes :)

wesoudshoorn:
Oh and maybe add "Ruby magic" as explanation?

tombruijn:
The normal API key is already on the list :)

I'll add the ruby magic one as well. 

Other suggestions are welcome!

wesoudshoorn:
Noticed that, just meant that maybe you can "cross-reference" them to make it absolutely clear that there are two and how they're different 👍

tombruijn:
Will do!

tombruijn:
Terminology updated with the last few missing topics. Please review, but also feel free to update the content yourselves. I've kept the content simple and high-level, not diving too deep in certain topics and referring to pages with more detailed explanations instead.

source/appsignal/terminology.html.md
```diff
@@ -0,0 +1,460 @@
...
+- [Samples](#samples)
+- [Stroopwafels](#stroopwaffles)
+- [Stroopwaffles](#stroopwaffles)
+- [Teams](#teams)
+- [Transactions](#transactions)
+- [User account](#user-account)
+
+## Agents
+
+AppSignal uses agents to monitor applications and hosts. These agents can be
```
matsimitsu: I think we call these extensions (ruby extension, elixir extension), the extension sends data to the agent.

jeffkreeftmeijer: I've been calling the Elixir extension an "integration", as it integrates in your app, and reports to the agent.

source/appsignal/terminology.html.md
```diff
@@ -0,0 +1,460 @@
...
+## Configuration
+
+The configuration is part of the agents running in applications. It tells the
+AppSignal [agents](#agents) what to instrument in an
+[application](#applications), which application it is and in which environment
+it's running.
+
+The AppSignal agents have multiple methods of configuration. The most common
+method of configuration is the usage of an `appsignal.yml` configuration file.
+The usage of environment variables is also recommended.
```
jeffkreeftmeijer: We'll need to mention the config file for Elixir here, eventually.

Mostly a note to self, here: Currently, it's stored in the application's main config file `config/config.exs`, but I'm quite sure this will change when we implement `mix appsignal install` (the installation task for the Elixir integration). It'll probably be something like `config/appsignal.exs`, but that might change.

source/appsignal/terminology.html.md
```diff
@@ -0,0 +1,460 @@
...
+
+For the configuration of the Ruby agent we recommend you read the
+[configuration topic](/ruby/configuration/index.html) to get started.
+
+## Environments
+
+Most [application](#applications) can be run in different modes. During
+development of an application other rules apply and errors are not usually
+shown in the form of a "500 internal server error" page like they do on
```
jeffkreeftmeijer: Nitpick: "in production" instead of "on production".

source/appsignal/terminology.html.md
```diff
@@ -0,0 +1,460 @@
...
+
+The first is about what libraries and frameworks the AppSignal agents support
+and the latter is about what services AppSignal.com integrates with.
+
+### Agent integrations
+
+AppSignal [agents](#agents) integrate with a variety of different libraries and
+frameworks specific for the programming language. The Ruby agent integrates
+with Rails, Sinatra and other available frameworks. The Elixir agent integrates
+with Phoenix.
```
jeffkreeftmeijer: The Elixir agent integrates with both Elixir and Phoenix.

source/appsignal/terminology.html.md
tombruijn: I'm okay with @jeffkreeftmeijer's suggestion 👍

tombruijn: Trying to rewrite this part and see a problem with the "Ruby integration" part. The Ruby gem also integrates with Rails and other frameworks. How would we call those things? I'm thinking that "library" is the best choice here and say we have "Rails integration".

And I'm thinking of using one name for the external integrations when used to name all of them: "Third-party integrations"

tombruijn:
Terminology updated. Please re-review @jeffkreeftmeijer @matsimitsu

PR #55, opened 2016-11-01
Reduce margin bottom for nested lists

tombruijn:
This change creates a more connected feel between nested lists and their parent list.

Example page: http://docs.appsignal.com/appsignal/contributing.html

## Before

image

## After

image

### DIFF
--- a/source/stylesheets/modules/_article.css.sass
+++ b/source/stylesheets/modules/_article.css.sass
@@ -41,6 +41,9 @@
     margin: 0 0 24px 20px
     list-style: disc
 
+    ul, ol
+      margin-bottom: 0
+
   ol
     list-style: decimal
 


wesoudshoorn:
Reducing margin is fine, but the example of the before / previous is actually not how we should probably use lists ;) It would be more fitting to have a <br> tag and just one level of list items.

Anyway, we can merge this!

tombruijn:
This is one example. I have more 😄 

Like a Table of Contents: http://docs.appsignal.com/api/markers.html

I'm not sure what you mean with the tag solution.

wesoudshoorn:
It seems github recognizes a `<br>`tag and creates a new line :D Put it in quotes, should make sense now.

tombruijn:
Yes, that would make more sense there

PR #56, opened 2016-11-01
Tweaks heading and paragraph spacing

tombruijn:
With more content and with more headings the usage of h3 and h4s was
confusing. Also decreased the spacing around the headings so that the
higher level headings create more spacing between sections than lower
level headings.

We should avoiding using a heading lower than h4.

In short this PR:

- Makes the distance between paragraphs smaller
- Gives h3 a bigger font size
- Tweaks the distance between headings and paragraphs.

## Before

image

## After

image

### DIFF
--- a/source/stylesheets/modules/_article.css.sass
+++ b/source/stylesheets/modules/_article.css.sass
@@ -14,25 +14,34 @@
     border-bottom: 1px solid $border
     padding-bottom: 10px
 
-  h3, h4, h5
-    @extend %h-small
+  h3, h4
     margin-bottom: 5px
+
+  h3
+    @extend %h-small
+
+  h4
+    @extend %h-tiny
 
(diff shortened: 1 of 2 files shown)


wesoudshoorn:
Looks good! I've been skimming the pages of the documentation and some of them are long and hard to see the structure of the page. I highly recommend cutting these pages up. Any time we use a H4, it means that the user has to keep 3 levels of context above it in mind. 

If you can't avoid a H4, using tables, lists or indenting of a section will help the user a lot.

tombruijn:
Yes, some of them are really long 😢 
I'm cutting them up as we speak, while I read the pages to see what documentation we do have.

I had the `h4` problem on the API Marker page (http://docs.appsignal.com/api/markers.html) where I can't really cut them up right now. I used more tables though to minimize it a little

PR #61, opened 2016-11-03
Add debugging AppSignal page

### LINKED ISSUE #39: 'Ways to debug AppSignal' page
Add a debugging page to the docs.

I run into this myself a lot, remembering commands, config options where to look for things

I think it can definitely be of help to people before they have to contact support (not saying you shouldn't!)

Things that should go on this page/these pages:
- [x] How can I see if AppSignal is running?
- Ways to debug AppSignal:
  - [x] run `appsignal diagnose`
  - [x] how to read the `appsignal.log` file
  - [ ] where does appsignal look for my config file?
  - [x] contact support

tombruijn:
Closes #39

### DIFF
--- a/source/support/debugging.html.md
+++ b/source/support/debugging.html.md
@@ -7,9 +7,24 @@ procedures to debug where a problem might come from.
 
 Since it's good to share, here are a couple of our procedures.
 
+## Table of Contents
+
+- [Diagnose](#diagnose)
+  - [Diagnose information](#diagnose-information)
+- [Configuration](#configuration)
+- [No data appearing](#no-data-appearing)
+- [Logs](#logs)
+  - [Logs contents](#logs-contents)
+  - [Log message breakdown](#log-message-breakdown)
+  - [Log location](#log-location)
+- [Is the agent running?](#is-the-agent-running)
+- [Other questions](#other-questions)
+- [Creating a reproducible state](#creating-a-reproducible-state)
+- [Contact us](#contact-us)
+
 ## Diagnose
 
-Our [Ruby agent](/ruby/index.html) ships with a built-in diagnose command line
+Our [Ruby agent](/ruby/index.html) ships with a built-in diagnose command-line
 tool that outputs information about the configuration of the AppSignal gem and
 environment it's running in. All this information can help in finding a
 potential cause of a problem.
@@ -43,14 +58,20 @@ The diagnose command will output the following data.
   - Hardware architecture
   - Operating system
   - Ruby version
-  - Container id - Docker/LXC
+  - Heroku detection - only on Heroku
+  - Container id - only in a Docker/LXC container
 - Configuration
+  - Environment - Outputs a warning if none is found.
   - See [Ruby configuration](/ruby/configuration/options.html) for all
     available options.
 - Push API key validation
   - Tests if the Push API key that's being used is a valid key at AppSignal.com.
 - Path information
   - Returns if all the required paths exist and if they are writable.
+  - `root_path` - application path. AppSignal will try to find a
+    `config/appsignal.yml` file here.
+  - `log_path` - path the log file is created. Is empty if no viable path could
+    be found.
 - Installation information
   - Something could have gone wrong during the installation of the AppSignal
     agent. This section outputs the `install.log` and `mkmf.log` (Makefile log)
@@ -69,14 +90,18 @@ instrumentation.
 ## Configuration
 
 The configuration is key. It's important, because without it we the AppSignal
-agent won't know which application it's instrumenting or in which environment.
+agent won't know which application it's instrumenting and in which environment.
+
+Using the [diagnose](#diagnose) information we see if we can find a problem
+with the configuration of the agent.
 
 The AppSignal agents have multiple methods of configuration. For the Ruby agent
 we recommend you read the [configuration topic](/ruby/configuration/index.html)
 to get started, specifically [minimal required
-configuration](/ruby/configuration/#minimal-required-configuration),
-[configuration load order](/ruby/configuration/load-order.html) and
-[configuration options](/ruby/configuration/options.html).
+configuration](/ruby/configuration/#minimal-required-configuration), the
+[configuration load order](/ruby/configuration/load-order.html) and the
+[configuration options](/ruby/configuration/options.html) if you're
+experiencing problems.
 
 ## No data appearing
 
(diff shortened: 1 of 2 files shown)


source/support/debugging.html.md
```diff
@@ -0,0 +1,201 @@
...
+---
+
+When a support request comes in at support@appsignal.com we have a couple
+procedures to debug where a problem might come from.
+
+Since it's good to share, here are a couple of our procedures.
+
+## Diagnose
+
+Our [Ruby agent](/ruby/index.html) ships with a built-in diagnose command line
```
jeffkreeftmeijer: > Our [Ruby agent](/ruby/index.html) ships with a built-in diagnose command line tool in that outputs as much information about the configuration of the AppSignal gem and environment it's running in. 

Broken sentence. How about this (just removed some words)?:

> Our [Ruby agent](/ruby/index.html) ships with a built-in diagnose command line tool that outputs information about the configuration of the AppSignal gem and environment it's running in.

source/support/debugging.html.md
```diff
@@ -0,0 +1,201 @@
...
+tool in that outputs as much information about the configuration of the
+AppSignal gem and environment it's running in. All this information can help in
+finding a potential cause of a problem. So usually this is the first thing we
```
jeffkreeftmeijer: > So usually this is the first thing we ask you to run in a support request. 

We can drop the "So" here:

> Usually this is the first thing we ask you to run in a support request. 

Or, even:

> If you open a support request, we'll usually ask you to run this first.

source/support/debugging.html.md
```diff
@@ -0,0 +1,201 @@
...
+- Push API key validation
+  - Tests if the Push API key that's being used is a valid key at AppSignal.com.
+- Path information
+  - Returns if all the required paths exist and if they are writable.
+- Installation information
+  - Something could have gone wrong during the installation of the AppSignal
+    agent. This section outputs the `install.log` and `mkmf.log` (Makefile log)
+    files.
+
+The following options needs to be correctly configured for AppSignal to start.
```
jeffkreeftmeijer: > The following options needs to be correctly configured for AppSignal to start.

"need" instead of "needs":

> The following options *need* to be correctly configured for AppSignal to start.

source/support/debugging.html.md
```diff
@@ -0,0 +1,201 @@
...
+the AppSignal agent installed.
+
+This sends a fake web request and error sample to the AppSignal servers in name
+of the application. This allows the AppSignal servers to process the data and
+create an application on AppSignal.com. This process is also used in the
+`appsignal install` command line tool to help with the installation process.
+
+## Logs
+
+When there's not an immediate problem found in the diagnose information the
```
jeffkreeftmeijer: > When there's not an immediate problem found in the diagnose information the next best thing is the AppSignal logs. 

"no" instead of "not", dropped "immediate", changed "the next best thing", "are" instead of "is":

> When there's no problem found in the diagnose information the next thing to check are the AppSignal logs.

source/support/debugging.html.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Logs location
+
+These logs are not written to your application's log files, but have their own
+separate `appsignal.log` file. Depending on the permissions of a host the agent
+will write to a different location.
+
+It will first try to create a `appsignal.log` file in an application's root
+directory. For Rails applications it will create a log file in the `log/`
+directory instead. If it has no luck creating a log file it will fallback on
```
jeffkreeftmeijer: > If it has no luck creating a log file it will fallback on the system's `/tmp` directory.

"fall back" instead of "fallback":

If it has no luck creating a log file it will *fall back* on the system's `/tmp` directory.

source/support/debugging.html.md
```diff
@@ -0,0 +1,201 @@
...
+- Since when does the problem occur?
+  - Was the agent updated?
+  - Were other libraries updated?
+- What do the AppSignal agent logs say with log level "debug"?
+- Is the [Operating System](/support/operating-systems.html) supported?
+- Is the application running inside a containerized system?
+
+## Creating a reproducible state
+
+When it's not immediate obvious what the cause of a problem could be, the next
```
jeffkreeftmeijer: > When it's not immediate obvious what the cause of a problem could be, the next step is to try and replicate the situation in the most minimal setup possible.

"immediately" instead of "immediate":

> When it's not *immediately* obvious what the cause of a problem could be, the next step is to try and replicate the situation in the most minimal setup possible.

Also, since we're quite a long way into debugging now, using "immediately obvious" is probably out of place. Instead, maybe we could do something like this?:

> If the steps described above don't fix the problem, a good next step is to try to replicate the situation in the most minimal setup possible.

source/support/debugging.html.md
```diff
@@ -0,0 +1,201 @@
...
+step is to try and replicate the situation in the most minimal setup possible.
+Create a test application with as little configuration as possible, load the
```
jeffkreeftmeijer: > Create a test application with as little configuration as possible, load the least amount of required libraries and other dependencies.

"and" instead of a comma:

> Create a test application with as little configuration as possible *and* load the least amount of required libraries and other dependencies.

source/support/debugging.html.md
```diff
@@ -0,0 +1,201 @@
...
+least amount of required libraries and other dependencies.
+
+With this reproducible example application we can start tracking down the cause
+of the problem. This means deep diving in the language specific agent, the
+C-extension and Rust agent. If there doesn't seem to be a problem, we
```
jeffkreeftmeijer: > If there doesn't seem to be a problem, we

Incomplete sentence. I don't know what you were going to say here, but I think the paragraph above is quite complete.

source/support/debugging.html.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Contact us
+
+[Contact us](mailto:support@appsignal.com), if you haven't already, and let us
+know what your problem is.
```
jeffkreeftmeijer: > [Contact us](mailto:support@appsignal.com), if you haven't already, and let us know what your problem is.

Sounds a bit harsh. How's this (maybe the "here to help"-bit is a bit too much)?:

> Don't hesitate to [contact us](mailto:support@appsignal.com) if you run into any issues while debugging the AppSignal client. We're here to help.

source/support/debugging.html.md
tombruijn: That's probably why I stopped 😅

tombruijn:
Updated. Thanks @jeffkreeftmeijer !

Ready for review
Let me know if anyone knows of more debugging steps they use.

source/support/debugging.html.md
```diff
@@ -0,0 +1,240 @@
...
+- Configuration `Environment` is set.
+- Configuration `name` is set.
+- Configuration `push_api_key` is set.
+- Configuration `active` is set to `true`.
+- The Push API key validates. An internet connection is required for this.
+
+## Configuration
+
+The configuration is key. It's important, because without it we the AppSignal
+agent won't know which application it's instrumenting and in which environment.
```
jeffkreeftmeijer: > The configuration is key. It's important, because without it we the AppSignal agent won't know which application it's instrumenting and in which environment.

Dropped the "we" before "the AppSignal agent":

> The configuration is key. It's important, because without it the AppSignal agent won't know which application it's instrumenting and in which environment.

source/support/debugging.html.md
```diff
@@ -0,0 +1,240 @@
...
+installation process.
+
+Note that the "demo" command has to be run in the directory of an application
+that has the AppSignal agent installed. The "demo" command was added in Ruby
+gem version `2.0.0`.
+
+## Logs
+
+When there's no problem found in the diagnose information the next thing we
+look at are the agent logs. The AppSignal agents create log files to output
```
jeffkreeftmeijer: > When there's no problem found in the diagnose information the next thing we look at are the agent logs.

Provided "we" is AppSignal, and we're giving debugging tips to "you" (the user), we should use "you" instead of "we" here:

> When there's no problem found in the diagnose information the next thing you can look into at are the agent logs.

Also, switching this around a little:

> When there's no problem found in the diagnose information,  the agent's logs are the next thing you can look into.

tombruijn: I tried to minimize the you/we-usage, because it's both for us (AppSignal) and new team members and possibly users that want to debug problems themselves.

Text updated

PR #62, opened 2016-11-04
New side navigation design

### LINKED ISSUE #57: Losing scroll position on refresh
When I refresh a page the scroll position resets and I get scrolled back up to the top of a page. Really annoying since I refresh the page a lot nowadays, updating the docs.

wesoudshoorn:
Trying to improve the docs and resolve the scroll issue by refactoring the layout a bit. 

## Things I am trying to improve
- Navigation is long and overview is hard in 1 look, toggling navigation items might help
- Refactor some CSS to make typography / header it consistent with the blog

## Some issues
- [x] Switching between pages in the documentation resets your opened / closed navigation items. Need a dev to look at this
- [x] I wrecked the responsiveness

Closes #57

### DIFF
--- a/source/_header_nav.html.haml
+++ b/source/_header_nav.html.haml
@@ -1,3 +1,0 @@
-%nav
-  %ul
-    %li= link_to 'Visit AppSignal.com', 'https://appsignal.com', :class => 'button small outline-white'
--- a/source/_side_nav.html.haml
+++ b/source/_side_nav.html.haml
@@ -1,7 +1,7 @@
 %nav.mod-side_nav
   %ul
     %li.section
-      %h2.open
+      %h2{:data => { :menu => "getting-started" } }
         Getting started
         = image_tag "arrow.svg"
       %ul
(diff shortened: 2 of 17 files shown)


tombruijn:
Ah I thought we decided to show everything always. So that a look through the nav would tell you exactly all the content we have. Is the list already becoming too long?

But yeah 👍  for collapsing content. When you don't use Ruby or Elixir you can collapse all the content for it.

Right now, some content is *gone*. What's your idea for the sub-levels under sections?

appsignal documentation 2016-11-04 12-15-52


We can save the collapsed/uncollapsed sections in a cookie with some JS.

wesoudshoorn:
Oh accidentally hid the subnav 😆 

**This would be how you land on a page by default. Current section open.**
image

**If you open up everything by default, the content already gets quit long.**
I might have said otherwise before, but seeing it like this does not make it easier to scan for me.
image

matsimitsu:
We should be able to detect an active sub-section from the url and open that part of the menu when navigating.

thijsc:
What is the status of this?

tombruijn:
@wesoudshoorn needs to fix the responsiveness and we need to fix opening the right topic/category block on load.

wesoudshoorn:
I made it responsive now @tombruijn @thijsc and the only thing we need to do is opening / closing the right sections. This is something a dev can do :)

tombruijn:
Added logic to toggle the navigation menus with javascript. Let me know what you think

PR #65, opened 2016-11-21
Update screenshots

### LINKED ISSUE #10: Update screenshots
Screenshots are outdated, and we should update them with automatically with Capybara. 

---

@tombruijn hijacking this issue:

Pages without screenshots, but desperatly need some:

- [ ] http://docs.appsignal.com/api/markers.html
- [ ] ...

tombruijn:
Fixes #10


tombruijn:
@wesoudshoorn  I create new screenshots with this task: https://github.com/appsignal/appsignal-server/pull/1856

But the UI has changes compared to the design used in the original screenshots. Maybe we'll have to do some JS/CSS tweaking to make certain elements the correct size before we take the screenshots.

Some screenshots are really off right now. To wide or too big in general, see this screenshot of the screenshots on http://localhost:4567/getting-started/new-application.html locally:

image


I also removed the spacing/border around the screenshots and added padding in the screenshots themselves for the box-shadow the elements now have.

What do you think?

wesoudshoorn:
If we are taking screenshots of elements I think we're best off to create a bit of CSS that makes the screenshots background and square and only take a screenshot of the element itself. We can then add shadows / borders in the docs to make them look nice in the text.

Does that make sense? For screenshots like "install appsignal gem" I would probably remove the title and button and only take a screenshot of the box itself :)

tombruijn:
> If we are taking screenshots of elements I think we're best off to create a bit of CSS that makes the screenshots background and square and only take a screenshot of the element itself.

I can tweak it a bit, but not all content will fit as well in a square box I think. But we'll see...

> We can then add shadows / borders in the docs to make them look nice in the text.

So I should disable the shadows in the screenshots so we can add them later? Or should I put back the border around the elements then?

> Does that make sense?

Hmm.. See my questions ;) 

> For screenshots like "install appsignal gem" I would probably remove the title and button and only take a screenshot of the box itself :)

Will update

tombruijn:
@wesoudshoorn updated the screenshots. Made them less wide overall. Some screenshots look better than others, but I think it's an improvement. What do you think?

tombruijn:
@wesoudshoorn based on feedback from call:

- Removed border-radius from elements in screenshots to avoid dealing with the fancy gradient background being visible under the rounded corners
- Use bigger screen for notifications screenshot
- Update webhook screenshot to new design

👍 /:-1: ?

wesoudshoorn:
Looks great @tombruijn !!

PR #66, opened 2016-12-07
Update CoffeeScript Object Loop

iamjohnford:
Since tags are really a hash/object `or` should be used to loop. `in` creates an array loop so `@tags` never gets populated.

### DIFF
--- a/source/front-end/error-handling.html.md
+++ b/source/front-end/error-handling.html.md
@@ -114,7 +114,7 @@ class @Appsignal
     @action = action
 
   tag_request: (tags) ->
-    for key in tags
+    for key or tags
       @tags[key] = tags[key]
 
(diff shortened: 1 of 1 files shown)


tombruijn:
Hi @iamjohnford, Thanks for the PR!

Did you mean `of` instead of `or`? `or` doesn't compile coffeescript for me 😞 

I think we meant this code instead:

```coffee
for key, value of tags
  @tags[key] = value
```

I'll fix it really quickly now.

tombruijn:
Fixed in https://github.com/appsignal/appsignal-docs/commit/7c788ec8fbc22202b88ffa849c8f95e72903c3fb

iamjohnford:
Hahah. Yep! Thanks for fixing that.

Issue #69, opened 2016-12-13
Document Elixir functionality


jeffkreeftmeijer:
- [x] Update configuration pages
- [x] Debugging AppSignal for Elixir #74 
- [x] Installation page. Mention manual setup but prefer install task
- [x] Integrating AppSignal in Elixir page, PR #88
  - See demo task for simple implementation
- [x] Custom instrumentation page updates #90 
  - See demo task for simple implementation
- [x] Custom metrics
- [x] Tagging PR: #89
- [x] Exception handling page PR: #87
  - `Appsignal.send_error`
  - like http://docs.appsignal.com/ruby/instrumentation/exception-handling.html
- [x] Update type definitions, as reported in https://github.com/appsignal/appsignal-docs/pull/85 PR: #86 
- [x] Things from: https://github.com/appsignal/appsignal-docs/issues/79

Make sure to update the docs in the Elixir page itself https://github.com/appsignal/appsignal-elixir so they don't contradict each other

PR #70, opened 2016-12-13
Add Elixir docs

roytomeij:
Move some of the Elixir documentation from Hexdocs to docs.appsignal.com, including an updated explanation on why we decided to use NIF's (this needs to be checked for technical inconsistencies).

Todo:

* [x] Add content to the "overview" page.

First setup for #69


roytomeij:
Please fix what's wrong, and try to stick to acknowledging that NIF's could be a bad idea but not in this case, and focus on customer benefits (battle tested among all customer).

thijsc:
Improved the NIF explanation.

source/elixir/why-nif.html.md
```diff
@@ -0,0 +1,15 @@
+---
+title: "Why a NIF?"
+---
+
+The AppSignal for Elixir integration relies on a NIF that calls out to a library that's written in [Rust](https://www.rust-lang.org) to do the heavy lifting. It's job is to receive data from the Elixir side, normalize and write it to a separate agent process over a Unix socket. The agent then sends the data to AppSignal.com periodically for further processing.
+
+While using NIF's has inherent risks, using one in this case has several advantages:
```
wesoudshoorn: "has inherent risks" -> can we add a section where we explicitly state these risks and what we did to minimize them? In what scenarios might a NIF become a problem and how do we handle those situation witch grace?

tombruijn:
I've updated the Ruby overview page to give a small intro to the different subjects of the Ruby documentation.

https://github.com/appsignal/appsignal-docs/compare/55e755179a022988e0b56afecdbd16e2394e3d48...602ce463f208ca623c125cee3736d8813be1488f

Maybe we can do the same for Elixir?

source/elixir/why-nif.html.md
roytomeij: That's a good one Wes. I addressed how me minimize those risks, but I didn't include what they are :-)

roytomeij:
We do need to add content to the `elixir/index` page, but I'm not sure what.

tombruijn:
I added a big ol' update. Split up pages and imported some more documentation from the hexdocs. Restructured it to match the Ruby docs.

@jeffkreeftmeijer Could you add an example to the Elixir decorator instrumentation in `elixir/instrumentation/index.html.md`? (I don't know how that works 🙊 )
That's the only thing missing now, I think.

PR #73, opened 2017-01-13
Update Elixir docs to show how to set params for filtering

tylerpearson:
Related to this PR https://github.com/appsignal/appsignal-elixir/pull/73. The code uses `Application.get_env(:appsignal, :config)[:filter_parameters]`, so `filter_parameters` needs to be set in `config`.


tombruijn:
Great! Thanks for the fix!
I copied this from the docs in the elixir project so we'll also have to fix it there.

PR #75, opened 2017-01-27
Fix heading anchors scroll position

tombruijn:
Currently when you scroll down to an anchor, by clicking an anchor link,
the heading you scrolled down to is hidden by the top header bar. That's
because it has a fixed positioned of being at the top.

This change uses a separate element for the anchor, which is positioned
above the heading element, so that the header doesn't move over it and
hide the heading you just scrolled down to.

This PR is already put online, in case you're testing it there, but is here for review.

## Before

scrolling down to "Agent" on the terminology page you can't see the heading you just scrolled to.

image

## After

scrolling down to "Agent" on the terminology page you can now see the heading

image

### DIFF
--- a/source/stylesheets/modules/_article.css.sass
+++ b/source/stylesheets/modules/_article.css.sass
@@ -22,9 +22,17 @@
     @extend %h-tiny
 
   h1, h2, h3, h4
+    position: relative
+
     a
       color: $font-heading
       text-decoration: none
+
+    .anchor
+      position: absolute
+      top: -($header-height + 20px)
+      pointer-events: none
+      user-select: none
 
   p + h2
     margin-top: 40px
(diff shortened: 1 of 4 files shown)


source/stylesheets/modules/_article.css.sass
```diff
@@ -22,10 +22,18 @@
...
   h1, h2, h3, h4
+    position: relative
+
     a
       color: $font-heading
       text-decoration: none
 
+    .anchor
+      position: absolute
+      top: -150%
```
wesoudshoorn: Why 150%? Isn't it smarter to do this based on the actual precise pixels we'd like to offset?

tombruijn: I've now based it on the `$header-height` (new variable) + some extra padding. Better?

PR #83, opened 2017-02-15
Update Ecto logger configuration instructions

arjan:
Keep Ecto.LogEntry there, otherwise queries are no longer outputted on the console (`Ecto.LogEntry` is the default logger configuration)


tombruijn:
Thanks! Merged with some additions

PR #84, opened 2017-02-15
Add GraphQL docs

matsimitsu:
Just a test to see if we can generate docs based on the graphQL schema, (so far I say we can)

- [ ] Clean up design
- [ ] Run through comments on GraphQL fields and types
- [ ] Add global text about GraphQL + how to use it

Part of https://github.com/appsignal/appsignal-server/issues/2333


source/graphql/query.html.md.erb
```diff
@@ -0,0 +1,23 @@
...
+
+Every GraphQL schema has a root type for both queries and mutations.
+
+The root type defines how GraphQL operations begin. It is the entry point to constructing GraphQL queries.
+
+<%- (graphql_query['fields'] || []).each do |field| %>
+##<%= field['name'] %>
+<%= field['description'] %>
+
+<%- if field['args'].any? %>
```
matsimitsu: Yes, that's erb inside markdown :P

thijsc:
One thing I'm wondering: People outside of AppSignal won't have access to the graphql schema file. Can they still run middleman without it?

thijsc:
Nevermind, we commit it :-)

thijsc:
I've updated the schema task and the schema itself. It now needs some design love and a general introduction. @wesoudshoorn shall we take a look at this?

wesoudshoorn:
@thijsc and myself have talked about designing this, but we decided that the content needed to be improved first before I could properly "design" this.

PR #85, opened 2017-02-23
correct types for values in options

Tuxified:
The current doc doesn't show the correct type of options. For example `Array<String>` gets rendered as `Array`, loosing important info (this also goes for the Ruby page). Next to that, it should be a list, arrays do exist in Erlang but basically nobody uses them. To make it clearer for Elixir devs you could use the Typespec notation: https://hexdocs.pm/elixir/1.0.5/Kernel.Typespec.html

In that case Boolean -> boolean, String -> String.t | nil, Array<String> -> list(String.t) etc

### DIFF
--- a/source/elixir/configuration/options.html.md
+++ b/source/elixir/configuration/options.html.md
@@ -51,7 +51,7 @@ Set this to `false` to disable [host metrics](/metrics/host.html).
 
 ## `APPSIGNAL_FILTER_PARAMETERS` / `:filter_parameters`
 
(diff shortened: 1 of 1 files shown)


tombruijn:
Welp.. Should have used code blocks for the types 😓

tombruijn:
Thanks for the PR! We'll look into updating the types to be more compatible with the Elixir Typespec, but for now we'll first fix the broken type definitions

PR #92, opened 2017-02-27
Add instruction for using Sidekiq standalone

thijsc:


source/ruby/integrations/sidekiq.html.md
```diff
@@ -7,6 +7,37 @@ for Ruby. It's also the processor we use to process jobs in AppSignal.
 
 Support for Sidekiq was added in AppSignal Ruby gem version `0.8`.
 
+## Using with Rails
+
 The AppSignal Ruby gem automatically inserts a listener into the Sidekiq server
-middleware stack if the `Sidekiq` module is present. No further action is
+middleware stack if the `Sidekiq` module is present if you use Rails. No further action is
```
tombruijn: Technically not only when using rails, right? but with any integration that automatically configures and starts AppSignal?

source/ruby/integrations/sidekiq.html.md
```diff
@@ -7,6 +7,37 @@ for Ruby. It's also the processor we use to process jobs in AppSignal.
...
+require 'appsignal'
+
+Sidekiq.on(:startup) do
+  # Initialize the logger
+  Appsignal.start_logger(Dir.pwd)
+
+  # Load config
+  Appsignal.config = Appsignal::Config.new(
+    Dir.pwd,
+    ENV['APPSIGNAL_ENVIRONMENT'],    # Set environment here
```
tombruijn: shouldn't this be `APPSIGNAL_APP_ENV`?

source/ruby/integrations/sidekiq.html.md
thijsc: Rails should be the only automatic integration that would configure Sidekiq out of the box.

source/ruby/integrations/sidekiq.html.md
thijsc: You're right, changing.

source/ruby/integrations/sidekiq.html.md
```diff
@@ -7,6 +7,37 @@ for Ruby. It's also the processor we use to process jobs in AppSignal.
...
+Sidekiq.on(:startup) do
+  # Initialize the logger
+  Appsignal.start_logger(Dir.pwd)
+
+  # Load config
+  Appsignal.config = Appsignal::Config.new(
+    Dir.pwd,
+    ENV['APPSIGNAL_APP_ENV'],        # Set environment here
+    :name   => 'Sidekiq standalone', # Set app name here
+    :active => true
```
tombruijn: Not sure if this should be here. The (active) config normally comes from the `appsignal.yml` file. Whether it is active or not should be based on the env, now it's always active

source/ruby/integrations/sidekiq.html.md
tombruijn: Yes, there can be scenarios where sinatra is active as well `require "appsignal/integrations/sinatra"`, but they should probably be separated and not loaded in the same process.

thijsc:
@tombruijn if Sinatra is active that wouldn't enable instrumentation for Sidekiq. This behaviour is as far as I know unique to Rails.

source/ruby/integrations/sidekiq.html.md
```diff
@@ -7,6 +7,36 @@ for Ruby. It's also the processor we use to process jobs in AppSignal.
...
+## Using standalone
+
+If you use Sidekiq without Rails some additional setup is required. Add this snippet to your Sidekiq config with the right environment and name:
+
+```ruby
+require 'appsignal'
+
+Sidekiq.on(:startup) do
+  # Initialize the logger
+  Appsignal.start_logger
```
tombruijn: this is odd, shouldn't this be called after the config is initialized? Otherwise it will start the logger based on the old/non-existent config.

thijsc: You're right, that should be the other way around nowadays.

PR #103, opened 2017-03-15
Use instrument/3 in the first instrumentation example

jeffkreeftmeijer:
Updated the first Elixir instrumentation example to use instrument/3 instead of instrument/4. 

⚠️ Requires https://github.com/appsignal/appsignal-elixir/pull/180 to be released first.


tombruijn:
👍  except that we should probably mention that `instrument/3` is only available in (assuming) 1.1.1 and up? That you need to use another way for < 1.1.1 for this to work

jeffkreeftmeijer:
Added a note with a code sample. The version is marked as TODO for now, we'll need to update that before merging.

PR #104, opened 2017-03-17
Add markers section

### LINKED ISSUE #93: Document deploy markers and APP_REVISON better
Now we have this page: http://docs.appsignal.com/push-api/deploy-marker.html
which doesn't describe the usefulness of the feature that well and only explains the API call

We also support the APP_REVISION env var which can be leveraged to route transaction samples per running instance of an app rather than an app-wide API "ping" that assumes all running instances of an app are, from that point on, running the new deploy

Related issue: https://github.com/appsignal/appsignal-elixir/pull/154

tombruijn:
Fixes #93 and https://github.com/appsignal/appsignal-elixir/issues/139

- [ ] Add images


tombruijn:
Added more content to this PR. Deploy markers and Custom markers are now documented.

I'd like to add more images to these pages, but I'm not sure how to automate that nicely. It requires some mouse hover input which is annoying to do. So for now I've added links to the blog which contains a nice little video. I can also add that here if people think that's better/also works.

PR #107, opened 2017-03-29
Update outdated text on Alpine Linux support

raytalks:
Alpine Linux has already been featured for both Ruby and Elixir. Text in docs was still about an experimental version and the issue mentioned has already been closed.

### DIFF
--- a/source/support/operating-systems.html.md
+++ b/source/support/operating-systems.html.md
@@ -10,7 +10,7 @@ most Linux distributions and macOS.
 
 ### Alpine Linux
 
-Alpine Linux support is currently in Alpha. You can try our experimental Alpine
+Alpine Linux support is currently in Alpha. You can try our Alpine
 Linux support by installing the `2.1.0.alpha.x` versions of the AppSignal gem.
 
 ```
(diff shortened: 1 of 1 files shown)


source/support/operating-systems.html.md
```diff
@@ -10,24 +10,20 @@ most Linux distributions and macOS.
 
 ### Alpine Linux
 
-Alpine Linux support is currently in Alpha. You can try our experimental Alpine
+Alpine Linux support is currently in Alpha. You can try our Alpine
 Linux support by installing the `2.1.0.alpha.x` versions of the AppSignal gem.
```
tombruijn: You don't need to install an alpha version anymore. "by installing version `2.1.0` or newer of the AppSignal gem" is fine.

source/support/operating-systems.html.md
```diff
@@ -10,24 +10,20 @@ most Linux distributions and macOS.
...
 
 ```
 gem "appsignal", "2.1.0.alpha.2"
```
tombruijn: here we should also remove `.alpha.2`

source/support/operating-systems.html.md
```diff
@@ -10,24 +10,20 @@ most Linux distributions and macOS.
...
 ```
 
 For the latest available alpha version see the full list on
```
tombruijn: also remove "alpha" here

PR #108, opened 2017-03-30
Fix wrong prefix for http-requests

p0wl:
http-requests should start with `process_action` and not `perform_action`, see https://github.com/appsignal/appsignal-ruby/blob/master/lib/appsignal.rb#L109


tombruijn:
hi @p0wl!


Thanks for the pr! Not sure how that came to be. Thank you for reporting it!

PR #109, opened 2017-04-07
Correct request method ♲

eval:
Untested - totally based on assumptions...


tombruijn:
Woops.. Looks like some copy paste or something happened there.

Thanks for the fix!

PR #110, opened 2017-04-11
Add new terminology words

### LINKED ISSUE #77: New terminology
I missed some words on the Terminology page:

Throughput = Total no. of requests sent through the controller in the selected timeframe.
Allocation = The number of times you're distributing memory in your application.
Impact = Duration impact. How much impact does this sample's duration have vs other samples.
Response time = Duration of a request. 
CPU = Hardware in computer that carries out the instructions of a program.
Queue time = How much time it takes for a request to be sent from X to Y (?)
Namespace = Name in Controller to define files and objects (?)

raytalks:
I added new terminology to the docs. Fixes #77.

### DIFF
--- a/source/appsignal/terminology.html.md
+++ b/source/appsignal/terminology.html.md
@@ -7,6 +7,7 @@ language of AppSignal.
 
 - [Agent](#agent)
 - [Alerting](#alerting)
+- [Allocations](#allocations)
 - [API](#api)
 - [API key](#api-key)
 - [Applications](#applications)
@@ -14,11 +15,13 @@ language of AppSignal.
 - [Blog](#blog)
 - [Changelog](#changelog)
 - [Configuration](#configuration)
+- [CPU](#cpu)
 - [Environments](#environments)
 - [Errors](#errors)
 - [Events](#events)
 - [Exceptions](#errors)
 - [Extension](#extension)
+- [Impact](#impact)
 - [Instrumentation](#instrumentation)
 - [Instrumentation events](#instrumentation-events)
 - [Libraries](#libraries)
@@ -26,11 +29,13 @@ language of AppSignal.
 - [Markers](#markers)
 - [Metadata](#metadata)
 - [Metrics](#metrics)
+- [Namespace](#namespace)
 - [Organizations](#organizations)
 - [Owners](#owners)
 - [Performance issues](#performance-issues)
 - [Push API](#push-api)
 - [Push API key](#push-api-key)
+- [Response time](#response-time)
 - [Ruby Magic](#ruby-magic)
 - [Samples](#samples)
 - [Stroopwafels](#stroopwaffles)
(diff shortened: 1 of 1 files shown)


source/appsignal/terminology.html.md
```diff
@@ -218,6 +234,24 @@ socket. To do so in a uniform way the libraries use an extension to the
 programming language they're written in. This extension is written in the
 C-language and installed when the language specific agent is installed.
 
+## Impact
```
jeffkreeftmeijer: Let's use the word "action" instead of "controller" in this section. Although the list is named "controllers", we're actually talking about an action on a controller here.

source/appsignal/terminology.html.md
```diff
@@ -170,6 +181,11 @@ The usage of environment variables is also recommended.
 For the configuration of the Ruby agent we recommend you read the
 [configuration topic](/ruby/configuration/index.html) to get started.
 
+## CPU
```
jeffkreeftmeijer: We're probably talking about CPU usage (also [CPU time](https://en.wikipedia.org/wiki/CPU_time)) instead of the CPU itself. Where was this mentioned?

source/appsignal/terminology.html.md
```diff
@@ -307,6 +341,12 @@ out causing the application to run much slower.
 
 Read more about [metrics](https://appsignal.com/for/metrics) on our tour page.
 
+## Namespace
+
+A class of elements (e.g. addresses, file locations, etc.) in which each element
+has a name unique to that class, although it may be shared with elements in
+other classes.
```
jeffkreeftmeijer: Correct, but in the context of AppSignal, we're talking about the namespaces our samples end up in. By default, these are "web" and "background", but you can use [custom namespaces](http://docs.appsignal.com/elixir/instrumentation/instrumentation.html#helper-namespaces) as well.

source/appsignal/terminology.html.md
tombruijn: I don't think we use "CPU" on its own anywhere, but it was one of the words in #77.

I don't think we need to mention what a CPU is, our demographic probably knows this, we can link to the wikipedia page if needed.
But maybe we can explain what the "CPU usage" graph is in this case.

tombruijn:
@jeffkreeftmeijer for context: this was a PR submitted by Rayta yesterday so that I, or someone else, could review and update the terminology which is not yet described completely. You caught it before I had time to look at it.

My plan was to update this PR, expand on the terminology and explain it in the context of AppSignal in collaboration with @raytalks

wesoudshoorn:
@tombruijn if you need to still work on this, shall we move it back to WIP?

tombruijn:
Sure, but Elixir Alchemy wasn't in the original request/PR. We can add that later in another PR :)

PR #117, opened 2017-07-03
Add Elixir Alchemy to Table of Contents

raytalks:
Add Elixir Alchemy to Terminology page, table of contents

### DIFF
--- a/source/appsignal/terminology.html.md
+++ b/source/appsignal/terminology.html.md
@@ -16,6 +16,7 @@ language of AppSignal.
 - [Changelog](#changelog)
 - [Configuration](#configuration)
 - [CPU usage](#cpu-usage)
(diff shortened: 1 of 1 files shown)


source/appsignal/terminology.html.md
```diff
@@ -16,6 +16,7 @@ language of AppSignal.
 - [Changelog](#changelog)
 - [Configuration](#configuration)
 - [CPU usage](#cpu-usage)
+- [Elixir Alchemy](## elixir-alchemy)
```
tombruijn: the `## elixir-alchemy` should be `#elixir-alchemy`.
That change will make it an anchor link to the right heading. Now it will create the following url: `/appsignal/terminology.html##%20elixir-alchemy` while it should be`/appsignal/terminology.html#elixir-alchemy`

PR #123, opened 2017-07-20
Add install instructions for CentOs & Fedora

jeffkreeftmeijer:


source/elixir/installation.html.md
```diff
@@ -33,6 +33,12 @@ apt-get install build-essential curl
 apk add --update alpine-sdk coreutils curl
 ```
 
+**CentOS / Fedora**
+
+```
+yum install gcc gcc-c++ make openssl-devel curl
```
tombruijn: Newer Fedora (25 and up) now use `dnf` as a package manager

Could you add another like for Fedora?

```
# Fedora 25+
dnf add gcc gcc-c++ make openssl-devel curl
```

PR #124, opened 2017-07-31
Update installation.md to point to latest version

ahmetabdi:


matsimitsu:
Thanks for your pullrequest :) 

`~> 1.0` should include all `1.x` releases ( `~> 1.0 == (> 1.0 && < 2.0)`). 
We'd rather keep the dependency on `~> 1.0`, so we don't have to update it on every release.

Could the reason for this pullreq be that there were issues with your dependencies (e.g. it used `1.1` instead of `1.2`)?

ahmetabdi:
Ah I see I misunderstood what it was actually doing didn't realise it included everything between 1.0-2.0 can close

PR #126, opened 2017-08-02
Plug instrumentation documentation

jeffkreeftmeijer:


source/elixir/integrations/plug.html.md
```diff
@@ -0,0 +1,89 @@
...
+instrumentation.
+
+This example app looks like the one we used before, but it has a slow function
+(aptly named `slow/0`) we'd like to add instrumentation for. To do that, we need
+to 
+
+1. `use Appsignal.Instrumentation.Decorators` to allow us to use AppSignal's
+   instrumentation decorators
+2. Decorate the function we want to instrument with the `transaction_event/0`
+   decorator
```
tombruijn: do we also want to recommend using the instrumentation helpers rather than only the decorators?

PR #132, opened 2017-08-30
Fix broken throughput anchor

NDuggan:


tombruijn:
Woops! My bad.. Thanks for the fix 👍

PR #135, opened 2017-10-03
Add 2fa page to documentation

jeffkreeftmeijer:
Closes https://github.com/appsignal/appsignal-server/issues/2491.


source/_side_nav.html.haml
```diff
@@ -49,6 +49,7 @@
             %li= link_with_active 'Manage owners', '/organization/team/owners.html'
             %li= link_with_active 'Manage teams', '/organization/team/teams.html'
         %li= link_with_active 'Billing', '/organization/billing.html'
+        %li= link_with_active 'Two-factor authentication', '/organization/two-factor-authentication.html'
```
tombruijn: This might belong better with a "User/Profile" section, rather than organization. You configure it on user-level.
The user section could also have things like pages about multiple email addresses (so people don't have to create multiple user accounts), user-level notifications, the other features on the security page, etc.
Can you add a "Your user account" section and move the page to that section?

source/organization/two-factor-authentication.html.md
```diff
@@ -0,0 +1,25 @@
...
+5. 2FA is now enabled. Don't forget to save the backup codes in a safe location. You can use them to sign into AppSignal if you ever lose your phone.
+
+## Disabling two-factor authentication
+
+1. Visit the [security page](https://appsignal.com/users/security) and click the "Disable two-factor authentication"-button.
+2. You can now remove AppSignal from your authenticator application on your phone
+
+## Regenerating backup codes
+
+We show the backup codes once after enabling 2FA for your account. You can generate new backup codes, but that will invalidate any existing ones.
```
tombruijn: @jeffkreeftmeijer what do you think of adding a small intro about what the backup codes are and why you might regenerate them?

> Backup codes are used to gain access to your account if you lose your phone or it breaks down. However, you may also lose your backup codes. If the latter occurs you can generate new backup codes.
>
> Normally we only show the backup codes once after enabling 2FA for your account. By generating new backup codes, it will invalidate any existing ones.

jeffkreeftmeijer:
Updated, please re-review.

PR #145, opened 2018-01-25
✍️ Recommend adding the `use Appsignal.Plug` line to top of endpoint

jeffkreeftmeijer:
Adding it directly before the router line is not needed anymore.


jeffkreeftmeijer:
Found it! This hasn't been true since we switched from a Plug to a wrapper arround `call/2` in https://github.com/appsignal/appsignal-elixir/commit/7e14e4ce5c97fe647dd2f475af7eec43e545cc35#diff-c62c8f3dc8d0f013058ee7dcf583307b, which was in 0.13.0. 😆

tombruijn:
Since it's been like this for a long time, no need to mention how it works in the old versions.

PR #157, opened 2018-04-03
Add (hidden) standalone agent documentation

matsimitsu:


source/appsignal/standalone-agent.md
```diff
@@ -0,0 +1,72 @@
...
+You can either start the agent manually
+
+```
+APPSIGNAL_AGENT_STANDALONE=true \
+APPSIGNAL_ACTIVE=true \
+APPSIGNAL_PUSH_API_KEY=<push api key> \
+APPSIGNAL_APP_NAME=AppSignal \
+APPSIGNAL_ENVIRONMENT=<environment> \
+APPSIGNAL_ENABLE_HOST_METRICS=true \
+APPSIGNAL_LOG=stdout \
```
tombruijn: wouldn't these env vars also be required to be underscored? since we're directly calling the agent here.

matsimitsu: ...yes yes they are!

matsimitsu:
@tombruijn fixed and force-pushed

tombruijn:
@matsimitsu great! I'm going to tweak it a little bit and then we can deploy it. Unless the "hidden" part is so hidden it shouldn't be?

matsimitsu:
@tombruijn nice! Just merge to master to deploy :) 

> Unless the "hidden" part is so hidden it shouldn't be?

I think it should be hidden as it doesn't describe an update path and we probably want to make a nice installer for it eventually.

PR #162, opened 2018-04-18
Add Heroku Host metrics page

matsimitsu:
Should we add it to the sidebar or keep it "hidden" for now?


tombruijn:
> Should we add it to the sidebar or keep it "hidden" for now?
Let's keep it hidden for now.

tombruijn:
Rewrote the page a bit, removed it from the sidebar and update its location. Still want to rename the "host.html" page to "host-metrics" later, but is beyond this PR's scope. Maybe even move it to the "application" section instead.

PR #174, opened 2018-05-08
Add cookie bar for docs

wesoudshoorn:
Same as https://github.com/appsignal/appsignal-server/pull/3214, but for docs

## Todo
- [x] Hide on click with javascript
- [x] Set cookie so people won't see it next time

### DIFF
--- a/source/javascripts/application.js
+++ b/source/javascripts/application.js
@@ -54,7 +54,7 @@ $(document).ready(function() {
     $(".logged-out").hide();
   }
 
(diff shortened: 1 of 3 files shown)


source/javascripts/application.js
```diff
@@ -54,6 +54,10 @@ $(document).ready(function() {
     $(".logged-out").hide();
   }
 
+  if (Cookies.get("appsignal_cookie_consent") == "true") {
```
tombruijn: https://github.com/appsignal/appsignal-server/pull/3214#discussion_r186961890

PR #177, opened 2018-05-17
Add section on processor parameter filtering

tombruijn:
Use a partial as the section is the same for all integrations.


tombruijn:
thanks!

PR #188, opened 2018-08-06
Remove all occurrences of "Regex" for metrics, only use wildcard. Part of #3403

matsimitsu:


tombruijn:
Let's merge when https://github.com/appsignal/appsignal-server/pull/3484 has been deployed. Moving to waiting for now.

PR #189, opened 2018-09-08
Fix typo in example code

timdiggins:
`e` isn't defined. You could replace it with `error` but bare raise is easier to read I think.

BTW, also, I think if you're going to use `rescue Exception => error` rather than `StandardError` -- then you should make a comment about what you're trying to catch in addition.

### DIFF
--- a/source/ruby/integrations/rails.html.md
+++ b/source/ruby/integrations/rails.html.md
@@ -16,7 +16,7 @@ begin
   require ::File.expand_path("../config/environment",  __FILE__)
 rescue Exception => error
   Appsignal.send_error(error)
-  raise e
+  raise
 end
 ```
 


tombruijn:
Hi @timdiggins! Thanks for the PR! 

What do you mean with "about what you're trying to catch in addition"? The `rescue Exception` behavior is the default in the AppSignal Ruby gem. It allows us to report all exceptions that could occur during start up without disrupting the normal crash behavior.

timdiggins:
@tombruijn I'm thinking of Exceptions like `SignalException` which probably you don't want to report. That said you do want to report (probably) `SyntaxError` and `LoadError` -- so maybe `rescue Exception => e` is correct. However so often `Exception` is the wrong thing to rescue that it would be good to add a comment why you're rescuing Exception (rather than `StandardError`, or just `rescue => e`)

https://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby

timdiggins:
specifically what I meant by "in addition" was "in addition to StandardError"

tombruijn:
Hey @timdiggins, just wanted to make sure you were talking about those exceptions. We're aware of how dangerous it is to rescue Exception. I've [written a blog post](https://blog.appsignal.com/2016/10/18/ruby-magic-exceptions-primer.html) about it as well 😅 and removed a lot of `rescue Exception` calls in the Ruby gem where they were not useful.

I think we do want to rescue Exception here, maybe not everyone wants to rescue SignalError specifically, that's why it's part of the docs and not part of the Ruby gem itself (yet): https://github.com/appsignal/appsignal-ruby/issues/383 It's open for discussion. Based on the number of requests we've seen, SignalException is something our users do want to track so I'll leave it as is. And we recommend [ignoring exceptions](https://docs.appsignal.com/ruby/configuration/ignore-errors.html) when they don't.

I'll have a look to see if we can maybe explain the bit of code a bit further and link to our blog to explain in more detail how rescuing works. Thanks for the suggestion 👍 !

timdiggins:
thanks for the link @tombruijn -- nice post!

PR #198, opened 2018-11-22
update sequel.html.md to fix 404

aksie:
fxied link to https://github.com/jeremyevans/sequel

### DIFF
--- a/source/ruby/integrations/sequel.html.md
+++ b/source/ruby/integrations/sequel.html.md
@@ -15,7 +15,7 @@ Read more about [`:instrument_sequel`][instrument-sequel].
 
 ### sequel-rails integration
 
-When using the [sequel-rails](https://github.com/jeremyevans/sequel) gem the instrumentation from the
+When using the [sequel-rails][sequel-rails] gem the instrumentation from the
 sequel-rails gem and AppSignal gem can [conflict when running
 migrations][instrumentation-issue].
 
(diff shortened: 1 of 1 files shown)


source/ruby/integrations/sequel.html.md
```diff
@@ -15,7 +15,7 @@ Read more about [`:instrument_sequel`][instrument-sequel].
 
 ### sequel-rails integration
 
-When using the [sequel-rails](sequel-rails) gem the instrumentation from the
+When using the [sequel-rails](https://github.com/jeremyevans/sequel) gem the instrumentation from the
```
tombruijn:
```suggestion
When using the [sequel-rails][sequel-rails] gem the instrumentation from the
```

This uses the existing link and points to the correct project

aksie: Hi, 
Robert told me it should link to https://github.com/jeremyevans/sequel

But I guess the point is it should not have been a link? / what do the square brackets do?

tombruijn: Sequel rails is another project than sequel itself. The square brackets reference the list of urls at the bottom of the file

matsimitsu:
So we've gone from round brackets to square brackets, correct?

aksie:
yes.
this wasn't meant to link to a page but to the list of urls in the footer.

PR #215, opened 2019-03-11
Add note to explain that argument needs to be uniq

jvanbaarsen:
The first argument for attach needs to be uniq within an umbrella app.
If this is not the case Telemetry will not gather any information for
that app.

### DIFF
--- a/source/elixir/integrations/phoenix.html.md
+++ b/source/elixir/integrations/phoenix.html.md
@@ -107,8 +107,7 @@ Telemetry.attach(
 )
 ```
 
--> **Note**: If you're running an umbrella app, the first argument for
-`attach` needs to be an unique name withing the umbrella app.
+-> **Note**: Telemetry's handler ID (the first argument to the :telemetry.attach/4 function) needs to be unique across your application. Make sure to use unique IDs when attaching the Ecto integration to multiple Repos.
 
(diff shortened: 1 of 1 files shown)


jeffkreeftmeijer:
This potentially isn't just a problem for umbrellas. I'd change this to something like:

>  **Note**: Telemetry's handler ID (the first argument to the `:telemetry.attach/4` function) needs to be unique across your application. Make sure to use unique IDs when attaching the Ecto integration to multiple Repos.

source/elixir/integrations/phoenix.html.md
```diff
@@ -107,6 +107,8 @@ Telemetry.attach(
 )
 ```
 
+-> **Note**: Telemetry's handler ID (the first argument to the :telemetry.attach/4 function) needs to be unique across your application. Make sure to use unique IDs when attaching the Ecto integration to multiple Repos.
```
jeffkreeftmeijer:
```suggestion
-> **Note**: Telemetry's handler ID (the first argument to the `:telemetry.attach/4` function) needs to be unique across your application. Make sure to use unique IDs when attaching the Ecto integration to multiple Repos.
```

jvanbaarsen:
@jeffkreeftmeijer Can you retrigger the build on Netlify?

jeffkreeftmeijer:
I can't find a way to do that without pushing an extra commit, as Netlify doesn't seem to have a retry button. Can we assume the backticks will work?

tombruijn:
Netlify has a retry button in their interface on a failed build

PR #217, opened 2019-03-20
Update parameter-filtering.html.md.erb

matsimitsu:
This suggests you have to create a file called config/appsignal.exs, while we mean place it in the current config file. 

https://app.intercom.io/a/apps/yzor8gyw/inbox/inbox/conversation/21263137876

### DIFF
--- a/source/elixir/configuration/parameter-filtering.html.md.erb
+++ b/source/elixir/configuration/parameter-filtering.html.md.erb
@@ -17,12 +17,12 @@ We support two ways of filtering parameters from being sent to AppSignal. Either
 
 ## AppSignal parameter filtering
 
-If you're not using Phoenix, or want to filter parameters without changing the
-Phoenix.Logger's configuration, you can set up filtered parameters in the
-AppSignal configuration file.
+If you're not using Phoenix, or want to filter parameters without changing the Phoenix.Logger's configuration, 
+you can set up filtered parameters in the generated config/appsignal.exs configuration file 
+(or in your app's main or application specific one in config/config.exs or config/{mix_env}.exs).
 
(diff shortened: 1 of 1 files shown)


source/elixir/configuration/parameter-filtering.html.md.erb
```diff
@@ -22,7 +22,7 @@ Phoenix.Logger's configuration, you can set up filtered parameters in the
 AppSignal configuration file.
 
 ```elixir
-# config/appsignal.exs
+# config/your_mix_env.exs
```
tombruijn: The installer creates a `config/appsignal.exs` file for the AppSignal config and loads it in all available envs with `import_config "appsignal.exs"` on install. Users can always merge it with their own env file, but that's up to the user.
I think this change isn't necessary?

jvanbaarsen: I believe the old description was more correct. I'm +1 with Tom.

jeffkreeftmeijer: You can install AppSignal without the installer, in which case you'd add it to one of your configuration files yourself. We could revert this and change the line above to:

>If you're not using Phoenix, or want to filter parameters without changing the Phoenix.Logger's configuration, you can set up filtered parameters in the generated `config/appsignal.exs` configuration file (or in your app's main or application specific one in `config/config.exs` or `config/{mix_env}.exs`).

(sorry, couldn't make this a suggestion because you didn't change this line first)

matsimitsu: Thanks! updated the text :)

PR #220, opened 2019-04-08
Update design_language submodule

tombruijn:
Includes new typography as present on AppSignal.com.
Bug fix for active tour item background color in navigation.


wesoudshoorn:
@tombruijn looks good, although I'd like to fix some things visually, but that's another issue ;-)

PR #227, opened 2019-07-09
Sourcemaps

matsimitsu:


config.rb
```diff
@@ -30,6 +30,18 @@ def canonical_url(page)
     "#{config[:protocol]}#{config[:host]}#{path}"
   end
 
+  def table_of_contents
```
tombruijn: ah I've thought about adding this a couple times now. May be good for simple Table of Contents, but there's also many pages that link to other pages from the ToC, which this does not support.

I would also prefer to use this through markdown. We can write a little helper for that in another PR.

matsimitsu: ok, I was already annoyed at maintaining the ToC after 10 minutes :P It seems quite error prone when making changes

tombruijn: It is! Wouldn't be the first time they broke 😬

tombruijn: Oh you already removed it 😅 
I can take a look to see where it would still be useful to have.

matsimitsu:
> @matsimitsu this should go live with the new JavaScript library? 

@tombruijn Yes! I've changed the merge base to the front end docs branch, so we don't accidentally deploy this before that's merged in.

PR #232, opened 2019-07-29
v1.0 Frontend docs update

xadamy:
Updates the documentation for the front-end integration with everything implemented for 1.0.


source/front-end/integrations/vue.html.md
```diff
@@ -1,37 +1,42 @@
 ---
-title: "AppSignal for JavaScript - Vue.js"
+title: "@appsignal/vue"
 ---
 
 ## Installation
 
-Add the `@appsignal/javascript` package to your `package.json`. Then, run `yarn install`/`npm install`.
+Add the  `@appsignal/vue` and `@appsignal/javascript` packages to your `package.json`. Then, run `yarn install`/`npm install`.
```
jeffkreeftmeijer:
```suggestion
Add the `@appsignal/vue` and `@appsignal/javascript` packages to your `package.json`. Then, run `yarn install`/`npm install`.
```

source/front-end/plugins/plugin-path-decorator.html.md
```diff
@@ -0,0 +1,14 @@
+---
+title: "plugin-path-decorator"
+---
+
+The `plugin-path-decorator` is a simple plugin that decorates all outgoing `Span`s with the current path. This is computed by reading from `window.location.pathname`.
```
jeffkreeftmeijer:
```suggestion
The `plugin-path-decorator` is a plugin that decorates all outgoing `Span`s with the current path. This is computed by reading from `window.location.pathname`.
```

source/front-end/span.html.md
```diff
@@ -0,0 +1,103 @@
...
+const span = appsignal.createSpan((span) => {
+  return span.setError(new Error("test error"))
+})
+
+appsignal.send(span)
+```
+
+The `send()` method is different to the `sendError()` method as it allows a `Span` to be passed as a parameter, which is either pushed immediately to the API, or in the case of a network error, added to the queue to be retried later.
+
+Once the `Span` is passed to `appsignal.send()`, any [Hooks]() are applied to the `Span` in the following order:
```
tombruijn: Missing path for the "Hooks" link here.

source/front-end/span.html.md
```diff
@@ -0,0 +1,103 @@
+---
+title: "Creating and Using a Span"
```
tombruijn: Could this page be considered the JavaScript version of the Ruby and Elixir custom instrumentation pages? (Not exactly the same.)
Maybe we can call it "Custom instrumentation"? So that it's consistent across integrations.

source/front-end/hooks.html.md
```diff
@@ -0,0 +1,48 @@
...
+  // scrub e-mail addresses
+  const { tags } = span.serialize()
+
+  return span.setTags({ 
+    text: tags.text.replace(/^[^\s@]+@[^\s@]+\.[^\s@]+$/, "[SCRUBBED]") 
+  })
+})
+```
+
+Plugins can add decorators and overrides, too. If you're looking to create a re-usable hook, that can be packaged in an `npm` module, for instance, we recommend utilizing [the plugin interface]() to register it.
```
tombruijn: missing path for the link here

xadamy: Fixed in 2149d1d.

source/front-end/span.html.md
xadamy: Fixed in 2149d1d.

source/front-end/span.html.md
xadamy: I'm not sure it is the same, I think "Custom instrumentation" implies quite a broad scope. What do others think?

tombruijn: I suppose it starts with spans, but over time we have more instrumentation tools?

xadamy: Let's add this at a later date then, I'm keen to get this closed and didn't hear anything from anyone else.

PR #233, opened 2019-07-29
Line label URL - correct path and anchor

wesgarrison:
Found this all the way at the bottom of https://docs.appsignal.com/metrics/custom.html#metric-tags. Just a path and anchor update.


jvanbaarsen:
@wesgarrison Thanks for the patch! 👍

PR #238, opened 2019-08-30
Add information about Json dependency

jvanbaarsen:
We removed the default dependecy on Poison, and made it optional. Add
something in the installation guide that tells users to install either
jason or posion.

### DIFF
--- a/source/elixir/installation.html.md
+++ b/source/elixir/installation.html.md
@@ -24,11 +24,16 @@ Before you can compile the AppSignal gem make sure the build/compilation tools a
 -> **Note**: If you're running an umbrella application, see also our [umbrella installation guide](/elixir/installation/umbrella.html).
 
 1. Start by adding `appsignal` to your list of dependencies in `mix.exs`.
+   AppSignal requires you to depend on Json encoder. You can pick between
+   `jason` and `poison`, our recommendation is to use `jason`.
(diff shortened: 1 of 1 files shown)


source/elixir/installation.html.md
```diff
@@ -24,11 +24,16 @@ Before you can compile the AppSignal gem make sure the build/compilation tools a
 -> **Note**: If you're running an umbrella application, see also our [umbrella installation guide](/elixir/installation/umbrella.html).
 
 1. Start by adding `appsignal` to your list of dependencies in `mix.exs`.
+   AppSignal requires you to depend on Json encoder. You can pick between
```
tombruijn:
```suggestion
   AppSignal requires you to depend on a JSON encoder. You can pick between
```

jeffkreeftmeijer:
This change will be included in the next minor release of the Elixir integration.

jvanbaarsen:
@jeffkreeftmeijer When will that be? Do we have a tracking issue for that release somewhere so we can add this PR to that todolist?

jvanbaarsen:
I've moved this to Waiting, since it is blocked by the release of the next release of the Elixir Integration

tombruijn:
Let's publish then when 1.11 is released next week.

PR #240, opened 2019-09-05
Add "Debugging Phoenix Releases"

jeffkreeftmeijer:
A quick guide for users having trouble with Distillery, or Elixir's releases feature.


source/elixir/debugging-releases.html.md
```diff
@@ -0,0 +1,33 @@
+# Debugging Phoenix releases
```
tombruijn: Do we think we have more to say about releases in the future?

Would this work better as part of the debugging guide in the support section?

tombruijn: Let's look at a better place to include this.

wesoudshoorn:
@jeffkreeftmeijer is this something that can go to "planned" or "wip"?

PR #241, opened 2019-09-05
Add Elixir 1.9 release binary instructions to diagnose page

jeffkreeftmeijer:
Elixir 1.9's release binaries have a different way to run internal commands, so the `appsignal_tasks` solution we've been using for Distillery won't work.


tombruijn:
Updated the CLI tool page here: https://github.com/appsignal/appsignal-docs/commit/1be7d84151ed497ea38663af9b97d2ebb76abfce

PR #248, opened 2019-09-25
Use the right templating example for link templates

dirkjonker:
The example says to use `{variable}`, while you need to use `%variable%`


matsimitsu:
I _knew_ I forgot something when making this change. Thank you for the pull! 🎉

dirkjonker:
@matsimitsu That was fast! You may also want to update the picture, that still contains the curlies.

matsimitsu:
@dirkjonker I've updated the screenshot as well, thanks again!

PR #250, opened 2019-09-26
Remove JS errors beta mentions

wesoudshoorn:
Fixes https://github.com/appsignal/appsignal-server/issues/4837

### DIFF
--- a/source/front-end/sourcemaps.html.md.erb
+++ b/source/front-end/sourcemaps.html.md.erb
@@ -1,5 +1,5 @@
 ---
-title: "Sourcemaps"
+title: "Sourcemaps <sup>Beta</sup>"
 ---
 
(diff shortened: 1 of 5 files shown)


source/front-end/sourcemaps.html.md.erb
```diff
@@ -1,14 +1,7 @@
 ---
-title: "Sourcemaps <sup>beta</sup>"
```
xadamy: Sourcemaps is going to remain in beta for the time being, can you confirm @matsimitsu ?

matsimitsu: Yes, we haven't had many successful usages of this. Out of beta, to me, implies it works well, which we currently don't know yet.

wesoudshoorn: It's back in beta!

PR #254, opened 2019-10-14
🚨New brand fixes, update header

wesoudshoorn:
**🚨Only merge after the https://appsignal.com/javascript/ landing page is live 🚨**

# Description
- Updates YAML files for navigation text
- Update new icons
- Add JS files


source/_header.html.haml
```diff
@@ -1,7 +1,7 @@
 %header.mod-header
```
jvanbaarsen: @wesoudshoorn What is the reason we are changing all the `"` into `'`?

wesoudshoorn: I copy paste this from the appsignal server project. Just changed the URLs to make it work here, this is the easiest way to keep the header in sync.

jvanbaarsen: @wesoudshoorn Ok, let's roll with it then :-) Reason I asked is that our codestyle is `"`, but I can see why in this case we don't use it 👍

PR #256, opened 2019-10-15
Breadcrumbs docs

### LINKED ISSUE #249: Add documentation for breadcrumbs
We currently have no documentation for the breadcrumbs feature, so this will need to be added before launch.

@matsimitsu Just checking to ensure I'm understanding how the fields should be used:

`action` = UI | Network | Navigation | Console
`message` = “The user clicked a button” | “HTTP 500 from http://blablabla.com” | “The user pressed back” | “A console.log message”

`metadata` = key/value metadata in <string, string> format

Does this seem okay to you?

xadamy:
Adds documentation for the breadcrumbs feature. Closes #249.


source/front-end/breadcrumbs.html.md
```diff
@@ -0,0 +1,39 @@
...
+import Appsignal from "@appsignal/javascript"
+
+const appsignal = new Appsignal({ 
+  key: "YOUR FRONTEND API KEY"
+})
+
+appsignal.addBreadcrumb({
+  category: "", // e.g. "UI", "Network", "Navigation", "Console"
+  action: "", // e.g "The user clicked a button", "HTTP 500 from http://blablabla.com"
+  metadata: {} // key/value metadata in <string, string> format
```
matsimitsu: maybe add the `message` param as well?

xadamy: @matsimitsu I thought we had dropped this in favour of the `action` or `category` field? Can you confirm?

matsimitsu: No message is still there, you can use either/and message and metadata.Message is for long stuff and metadata for key/value things (or at least that's how it's implemented in the view):)

xadamy: > you can use either/and message and metadata

So you mean you can use either `message` or `metadata`, and both at the same time?

matsimitsu: Correct!

xadamy: Thanks! 👍

source/front-end/breadcrumbs.html.md
```diff
@@ -0,0 +1,52 @@
...
+|  metadata  |  Object<string, string>  |  (optional) An object of metadata related to the event  |
+
+## Plugins
+
+There are a number of [plugins](#plugins) available to enable automatic collection of breadcrumbs, and more will be added as time goes on:
+
+- [`@appsignal/plugin-breadcrumbs-console`](https://github.com/appsignal/appsignal-javascript/tree/develop/packages/plugin-breadcrumbs-console)
+- [`@appsignal/plugin-breadcrumbs-network`](https://github.com/appsignal/appsignal-javascript/tree/develop/packages/plugin-breadcrumbs-network)
+
+Have a good idea for a breadcrumbs plugin?
```
matsimitsu: I feel like this sentence isn't finished yet :)

source/front-end/breadcrumbs.html.md
```diff
@@ -29,9 +33,18 @@ appsignal.addBreadcrumb({
...
+| ------ | ------ | ----- |
+|  category  |  String  |  Category to label the event under  |
+|  action  |  String  |  Contextual information related to the event  |
+|  message  |  String  |  (optional) A log message or other string to send to AppSignal  |
+|  metadata  |  Object<string, string>  |  (optional) An object of metadata related to the event  |
+
 ## Plugins
 
-There are a number of plugins available to enable automatic collection of breadcrumbs, and more will be added as time goes on:
+There are a number of [plugins](#plugins) available to enable automatic collection of breadcrumbs, and more will be added as time goes on:
```
wesoudshoorn: All our headings get automagic links, so the heading above should already be linkable through `#plugins`. Think we can remove this inline anchor?

source/front-end/breadcrumbs.html.md
xadamy: Indeed, fixed!

PR #258, opened 2019-10-18
Add our new Designlanguage

jvanbaarsen:
Now that we have a good sense on how the new design language should work it is time to start adding it to our projects. The first will be the docs; The reason for this is that it is a small contained project with not that much CSS.


postcss.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  plugins: [
+    require("postcss-import"),
+    require("tailwindcss")("./source/assets/design_language/tailwind.config.js"),
+    require('postcss-preset-env')({
+      autoprefixer: {
+        flexbox: 'no-2009'
```
wesoudshoorn: What does this do?

source/assets/stylesheets/application.css
```diff
@@ -0,0 +1,26 @@
+@import "tailwindcss/base";
+@import "tailwindcss/components";
+@import "tailwindcss/utilities";
```
wesoudshoorn: If I am correct, we should load the `utilities` _after_ our components. 

Since all our utilities _and_ our components are mostly a singles class such as `.c-button` and `.bg-red`, the order of these in the CSS define which are used.

In our current setup, we'd be unable to overwrite our components with utility classes. Since the `.c-button` is declared later in our document than `.bg-red`, and the specificity is the same, it'd ignore the `bg-red`.

source/assets/main.js
```diff
@@ -1,2 +1,8 @@
+// New css
```
wesoudshoorn: We might want to rename this to "tailwind" or "design-library" or something to make this clearer?

wesoudshoorn: Maybe we should also consider the order of this? We'd probably want Tailwind stuff to overwrite older stuff, if there's conflict?

postcss.config.js
jvanbaarsen: I actually copied this from Gatsby I believe. I'd need to do some investigation on what is actually means

source/assets/stylesheets/application.css
jvanbaarsen: Hm, I would need to recheck the docs, because previously this was not the case. You'd load all the Tailwind stuff first, and then you would define your own things.

source/assets/main.js
jvanbaarsen: My preference would be to rename the old thing, and keep the new one named as `application.css` (Since that is a bit of a pattern in our apps that application.css is the main entry point). So we could rename the old one to `application-legacy.css.sass` or something like that.

jvanbaarsen:
@wesoudshoorn Yes, the goal is to rebuild. But the first step is to add this and see what breaks. If the answer is nothing, I want to have this running in prod as soon as possible. That way we can ship the rebuild in small increments instead of one massive PR.

source/assets/main.js
jvanbaarsen: I renamed the old one to `application-legacy.css.sass`

jvanbaarsen: I renamed this to `application-legacy.css.sass`

source/assets/stylesheets/application.css
jvanbaarsen: @wesoudshoorn Good catch this one! I followed the Tailwind docs, and they actually direct you towards putting your own things underneath the utilities. This is not something we want to do. I'll make this change!

jvanbaarsen:
I'm noticing errors on the fonts. So hold off on the merge before I have that fixed.

wesoudshoorn:
> I'm noticing errors on the fonts. So hold off on the merge before I have that fixed.

I found this as well. Because we're loading the _old_ CSS after the new CSS, I think the reset.css is overwriting some of the fonts.

jvanbaarsen:
@wesoudshoorn I actually already changed that around. My guess is that there is something going wrong with Webpack and picking up the fonts.

jvanbaarsen:
Interesting, the fonts are even being loaded when the legacy css is not included. It is almost as if it is loaded via JS.

jvanbaarsen:
Well, this makes sense :P

This was hardcoded in the design language:
```
  src: url('/src/design-language/fonts/39F5F8_0_0-eot.eot');
  src: url('/src/design-language/fonts/39F5F8_0_0-eot.eot?#iefix') format('embedded-opentype'),
    url('/src/design-language/fonts/39F5F8_0_0-woff2.woff2') format('woff2'),
    url('/src/design-language/fonts/39F5F8_0_0-woff.woff') format('woff'),
    url('/src/design-language/fonts/39F5F8_0_0-ttf.ttf') format('truetype');
```

jvanbaarsen:
This is actually kinda crappy to solve :( Somehow Webpack doesn't really like relative paths. So this is going to be interesting to solve. We might need to move the font declarations towards the projects itself.

jvanbaarsen:
Wes and I discussed a couple of things in a call last week, as a result of that call I changed the following things:

* Renamed the new design-language repo to design-system
* Changed the application.css files after what they have, so the old system is called `design-language-legacy.css.sass` and the new system is called `design-system.css`.
* The submodule is renamed to `design-system`
* Moved the `fonts.css` file to the project where the design-system is included and made sure this is explained in the README.

postcss.config.js
jvanbaarsen: Ok, I looked this stuff up, the flexbox option is something that we set in the `autoprefixer` package and let autoprefixer know that we don't need any prefixing for browsers older then IE11.

Stage 3 means we want to use stage3 css options: https://cssdb.org/#staging-process

PR #259, opened 2019-10-18
Add "Elixir Mix Tasks"

jeffkreeftmeijer:
After a question in support; here's how to add instrumentation to a Mix task.


source/elixir/instrumentation/mix_tasks.html.md
```diff
@@ -0,0 +1,64 @@
...
+    Appsignal.Nif.stop
+  end
+end
+```
+
+## Measuring performance
+
+To measure performance in your task, start and stop a transaction manually and add events using the instrumentation helpers, like described in the [custom helper instrumentation](https://docs.appsignal.com/elixir/instrumentation/instrumentation.html#helper-transactions).
+
+```
```
jvanbaarsen:
```suggestion
```elixir
```

PR #261, opened 2019-10-29
Update Puma docs for new plugin solution

tombruijn:
Explain why the plugin is needed and what kind of additional
configuration may be required.

Documentation for https://github.com/appsignal/appsignal-ruby/pull/570
Part of https://github.com/appsignal/appsignal-ruby/issues/502


tombruijn:
Moved PR to waiting until a new gem version is released.

PR #272, opened 2019-11-19
Remove old css

jvanbaarsen:
We still had a bunch of dependencies in the project that we actually no longer need. This removes the following:

* Unused packages (No more nodeSass)
* Old CSS
* All the font files (They now live in the design system)


jvanbaarsen:
I'm gonna merge this in so it triggers a build on the other PR.

PR #273, opened 2019-11-19
Fix sidebar opening

jvanbaarsen:
Currently the sidebar didn't remember which items were opened and which
were closed. This fixes that, and also converts all that logic over to
plain old javascript.


jvanbaarsen:
Mergin this one in so it can deploy to preview branch

PR #274, opened 2019-11-19
Update plugin-path-decorator.html.md

atonse:
Update the docs to clarify that you do need to install a separate node package.


tombruijn:
Hi @atonse,

Thanks for the PR!

@xadamy can you review and merge if approved?

xadamy:
Good catch, thanks @atonse! 🎉

PR #277, opened 2019-12-03
Split the uninstall guide per language

tombruijn:
Split the uninstall guide for every language we support so the
installation instructions are specific to the language and we don't
confuse people with mentions of other tools.

Add it the bottom of the installation page so it's easier to find,
grouped with the guide to install it.

Related to https://github.com/appsignal/appsignal-ruby/issues/574

### DIFF
--- a/source/front-end/installation.html.md
+++ b/source/front-end/installation.html.md
@@ -2,6 +2,19 @@
 title: "Installing AppSignal for JavaScript"
 ---
 
+## Table of Contents
+
+- [Installation](#installation)
+  - [Requirements](#requirements)
+  - [Configuration](#configuration)
+      - [Manual configuration](#manual-configuration)
+  - [Run your application!](#run-your-application)
+  - [Optional: Add Phoenix instrumentation](#optional-add-phoenix-instrumentation)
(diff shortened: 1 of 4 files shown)


source/front-end/installation.html.md
```diff
@@ -41,3 +54,14 @@ Currently, we have no plans to supply a CDN-hosted version of this library.
 This package can be used in any ECMAScript 5 compatible browser. We aim for compatibility down to Internet Explorer 9 [(roughly 0.22% of all browsers used today)](https://www.w3counter.com/globalstats.php). All browsers older than this can only supported on a “best effort” basis, and full functionality cannot be guaranteed.
 
 When developing, don’t forget to check browser support on [Can I Use?](https://caniuse.com/) and the [ES6 Compatibility Table](https://kangax.github.io/compat-table/es6/), and provide the appropriate polyfills or fallbacks. **In a small percentage of browsers, a `Promise` polyfill may be required to use this library.**
+
+## Uninstall
```
tombruijn: @xadamy can you confirm these are clear enough instructions to uninstall the JavaScript packages?

tombruijn: thanks! I'll add it as an option and merge the PR.

PR #286, opened 2020-01-17
Update data life cycle page and document time

tombruijn:
## Update data life cycle page

Update to the new integrations structure using the agent. I think it was
still taking about how Ruby gem version 0.x worked.

Rename the request lifecycle page to be about "data" as a more generic
term to also cover background jobs, custom transactions and custom/host
metrics.

This change also describe how Anomaly detection processing is done as
this question comes up in support every once in a while. This is a
simplified version leaving out a lot of details to keep the process
understandable.

An entry detailing how to set up NTP, or at least a separate section
detailing the problem which we can refer to from other pages as well,
will follow later.

## Add About time page

Describe how AppSignal works with times and to set up NTP on an app
servers, by linking to way more complete guides than what I can write.


source/appsignal/data-life-cycle.html.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "The life cycle of app data"
```
tombruijn: This page replaces https://docs.appsignal.com/appsignal/request-lifecycle.html
and is the important part of the PR. It details how anomaly detection processing works (and updates how sample and metrics processing works) so that we can refer to it when we get see this problem again in support.

@thijsc @matsimitsu can you verify it's broadly matches how the processing works. It doesn't need to be very detailed. It think this is as detailed as we want to go.

PR #293, opened 2020-01-30
Add section about namespaces vs applications on the "applications" page.

### LINKED ISSUE #185: Document how to use namespaces to report data from multiple apps
If you have multiple apps, but don't want to show them as separate apps in AppSignal.com, you can use namespaces.

Use a [namespace](https://docs.appsignal.com/application/namespaces.html) per application.

For example, in our AppSignal app we have a "processor" namespace that reports data from a different app, but is visible in AppSignal.com in the same app, just a different namespace.

matsimitsu:
Explains the up/downsides of each option and links to the corresponding
section for each choice (namespace or application).

I chose to not create a separate page for this, as it's not a common request. I'm actually not entirely sure we should mention this option at all, since setting up apps as namespaces requires quite a bit of configuration (before filter and overriding namespaces for background jobs for example). 

Maybe I should explain that a bit more in the namespace section? wdyt @tombruijn 

fixes #185

### DIFF
--- a/source/application/index.html.md
+++ b/source/application/index.html.md
@@ -24,6 +24,22 @@ A list of Applications appears on the [Application index] and in the application
 
 To add a new Application to AppSignal, please follow the [add a new application setup wizard](/application/new-application.html).
 
+When adding a new application you'll have to decide if you would like a separate application, or a new namespace for an existing applicaiton
+
+There are a couple of advantages and downsides for both options described below:
+
+### Separate applications
+
+The most common setup is to create a [new application](new-application.html) for each "app" in your ecosystem. E.g. if you have microservices, each microservice would be it's own app. This allows owners to cusomize who has access to each application and teams are able to setup their own notification settings and integrations.
+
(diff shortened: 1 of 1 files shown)


tombruijn:
> setting up apps as namespaces requires quite a bit of configuration (before filter and overriding namespaces for background jobs for example).

This was the reason I held off on documenting this. We should think about adding a `default_namespace` option to the integrations if we really want to offer this as a solution.

matsimitsu:
> This was the reason I held off on documenting this. We should think about adding a default_namespace option to the integrations if we really want to offer this as a solution.

Good point, maybe we shouldn't document it at all for now, since it's really only useful for our own very specific setup (a custom AppSignal integration that only performs things in a single namespace/area). I don't think a lot of customers have this specific scenario at the moment.

tombruijn:
> I don't think a lot of customers have this specific scenario at the moment.

I don't think so either. But maybe they do and can't find it now as there are no docs. But let's not document it as a solution until it's a convenient solution to use.

Let's close it for now?

matsimitsu:
Ok!

PR #297, opened 2020-01-30
Add documentation page for incident notification settings.

### LINKED ISSUE #180: Document notification and incident reopening configuration
> [the incident] should automatically reopen when it's set to "every occurrence". Do note there's a 5 minute delay for this option, otherwise we'd spam your mailbox/chat like crazy if you'd have 20 of these per second.

As explained in: https://app.intercom.io/a/apps/yzor8gyw/inbox/inbox/540709/conversations/16741968846 (private link)

Would also help if this was better in the UI, every place the notification settings are configurable:

- per incident
- per app & namespace
- per organization & namespace

matsimitsu:
Add page about incident-level notification options:

* how they work
* how they are triggered
* how defaults work

fixes #180


source/application/incident-notification-settings.html.md
```diff
@@ -0,0 +1,71 @@
...
+You can also set the [notification setting defaults](https://appsignal.com/redirect-to/organization?to=admin/notifications/edit) on an organization level. We allow you to set defaults for each namespace detected in any of the applications in your organization.
+
+
+![Performance incident notification defaults](/assets/images/screenshots/performance_incident_notification_defaults.png)
+
+## Notification settings inheritance.
+
+The notification settings bubble up from the incident settings > namespace settings > organization settings, if the settings aren't changed the defaults are used for the namespace or organization.
+
+* For an incident without any namespace overrides, the incident will use the organization settings.
```
tombruijn: I'm editing the page now, so don't make any changes yourself, but I noticed this is incorrect. We apply organization notification defaults upon creation of an app. Only then are they used. They don't trickle up. Can you confirm @matsimitsu

matsimitsu: Is that so? But we show all namespaces on the org-level screen, if those are only applied to new apps, how does that work exactly?

matsimitsu: hmmm some things also seem te be lost, notification options on the incident model is `:default` by default, which we used to show in the dropdown (we showed `(default)` behind the default option). 

It falls back to `site.exception_notification_defaults_for` when `default` is the option.
This does a query on its embedded `exception_notification_defaults` array.

It does not bubble up further indeed.

Site defaults gets filled on created from organisation.

This means that if you set organisation-level defaults for a "custom" namespace, and that namespace is not present during site create, it will never be used.

tombruijn: > This means that if you set organisation-level defaults for a "custom" namespace, and that namespace is not present during site create, it will never be used.

I think it's still used. Just not shown as the namespace isn't present in the app. Just tested creating a new app and it inherits the organization settings, they're just not visible.

Let's create a new issue for this in the server repo. Something is wrong if we're having trouble understanding it 😓

matsimitsu: Ahh right, so on site it's a "snapshot" of whatever org was on that moment. That seems fine, as long as we explain it. And incident defers to site if it's' not overridden.  Lets document the current state, so we have something to point to and maybe add a notice on the organisation settings to tell it's a snapshot and link to this page?

Shall I change it or will you?

tombruijn: I'm already updating it. You can review in a couple of minutes. Did quite a few changes.

tombruijn:
@matsimitsu I "tweaked" the page a bit. Please review :)

PR #301, opened 2020-02-25
Add note explaining instrumenters are no longer needed

jeffkreeftmeijer:
From 1.12 onward, the instrumenter doesn't need to be configured anymore ([as explained in the release blogpost](https://github.com/appsignal/appsignal-blog/pull/478/files)). Thus, the step could be removed completely.

However, I'm hesitant to remove this altogether, as we use this document to check users with existing setups. Instead, I'd like to propose adding a note here and removing the whole section once this version has been out for a while.

### DIFF
--- a/source/elixir/integrations/phoenix.html.md
+++ b/source/elixir/integrations/phoenix.html.md
@@ -45,7 +45,7 @@ endpoint.
 
 ## Phoenix instrumentation hooks
 
--> **Note**: From AppSignal for Elixir package version `1.12.0` onward, manually configuring Phoenix instrumentation hooks is no longer needed and this step can be skipped.
+-> **Note**: From AppSignal for Elixir package version `1.12.0` onward, manually configuring Phoenix instrumentation hooks is no longer needed and this step can be skipped for apps running Phoenix 1.4.7 and up.
 
 Phoenix comes with instrumentation hooks built-in. To send Phoenix'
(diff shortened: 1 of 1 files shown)


source/elixir/integrations/phoenix.html.md
```diff
@@ -45,6 +45,8 @@ endpoint.
 
 ## Phoenix instrumentation hooks
 
+-> **Note**: From AppSignal for Elixir package version `1.12.0` onward, manually configuring Phoenix instrumentation hooks is no longer needed.
```
tombruijn:
```suggestion
-> **Note**: From AppSignal for Elixir package version `1.12.0` onward, manually configuring Phoenix instrumentation hooks is no longer needed and this step can be skipped.
```

Make it more clear what the next step are (not).

tombruijn: Isn't this only the case for some new Phoenix versions? Like 1.5 and up?

jeffkreeftmeijer: This is the case for versions 1.4.7 and up. Including the instrumenters on 1.4.x anyway won't negatively affect instrumentation. On 1.5.x, including the instrumenters won't do anything and throw a warning.

I've updated the comment to recommend not adding them from 1.4.7 onward.

PR #318, opened 2020-05-20
Fix Stimulus link on front-end integrations page

conarro:


tombruijn:
Hi @conarro, Thanks for the PR! That's indeed broken 😄

PR #319, opened 2020-05-22
Update parameter_filtering.rb initializer

richardvenneman:
I like that you're demonstrating putting the parameter filtering in an initializer, however we can't directly access `config` here.


tombruijn:
Thanks for the PR @richardvenneman! `config` is not available there as you say.

PR #327, opened 2020-07-22
Upgrade to latest version of DesignSystem

jvanbaarsen:
This includes fontawesome pro, for this you need to use the auth token
that can be found in 1password.


tombruijn:
The only "token" I found in 1password was a "license key". Which gives me a 401 Unauthorized
Under what entry should the token be?

jvanbaarsen:
@tombruijn Somehow that was the wrong key. I updated it to the correct one 👍

README.md
```diff
@@ -14,14 +14,14 @@ Run the following commands:
 ```sh
 git submodule init
 git submodule update
-yarn install
+FORTAWESOME_TOKEN=find_token_in_1password yarn install
```
tombruijn, line 17: This does prevent any non-AppSignal people to install the package. They can't check out the repo and test the changes they make to the docs. I'd like to keep it low barrier to allow this though. Can we make the package optional for example?

README.md
```diff
@@ -14,14 +14,14 @@ Run the following commands:
...
 bundle install
 ```
 
 ## Usage
 
 ```sh
-bundle exec middleman
+FORTAWESOME_TOKEN=find_token_in_1password bundle exec middleman
```
tombruijn, line 24: The key is necessary here, but doesn't need to be a valid key.
Can we omit the check here or provide default dummy value?

I added this line to `config.rb` and that worked

```ruby
# config.rb
ENV["FORTAWESOME_TOKEN"] ||= "DUMMY_VALUE"
```

Otherwise I'd a `dotenv` approach so that you can store the key in a `.env` file and it's loaded that way instead of from the terminal session.

README.md
jvanbaarsen: I don't think we can. @xadamy Do you know if there is a way to make a package optional?

README.md
jvanbaarsen: I'll look into this.

README.md
xadamy: @jvanbaarsen Yes and no! It's complicated because JavaScript 🤷 I'll ping you on slack later and we'll talk about it

jvanbaarsen: 👍

xadamy:
@jvanbaarsen Good news! I think I have this at the point where it's not "fixed" per se, but is what I'd call "acceptably broken" in that it doesn't disrupt anyone who wants to make changes anymore.

I committed 26b806c, which has the following changes:

- `@fortawesome/fontawesome-pro` is now in `optionalDependencies`. Unfortunately, this seems to break `yarn`.

```
[3/4] 🔗  Linking dependencies...
warning "@appsignal/design-system > @tippyjs/react@4.1.0" has unmet peer dependency "react@>=16.8".
warning "@appsignal/design-system > @tippyjs/react@4.1.0" has unmet peer dependency "react-dom@>=16.8".
error An unexpected error occurred: "ENOENT: no such file or directory, open '/Users/adamyeats/Library/Caches/Yarn/v6/npm-@fortawesome-fontawesome-pro-5.14.0-0fef6462ad9450211cfbf6ce3f6e9bb8353a4982-integrity/node_modules/@fortawesome/fontawesome-pro/.yarn-metadata.json'".
info If you think this is a bug, please open a bug report with the information provided in "/Users/adamyeats/Code/appsignal-docs/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

`npm install` works as expected, though. However, you'll need to change the workflow slightly for this repo, so that a user would have to add their own `.npmrc` with a token to the directory to make the fontawesome stuff work, with all the config stuff inside. 

- In `webpack.config.js`, I'm using the `IgnorePlugin` to prevent compilation of the fontawesome CSS if the fontawesome token isn't available. This means that Webpack won't break, but you _will_ still see an error in the console in the browser. This at least makes it work.

Feel free to revert this change if it's unacceptable, but it's the best I can do without scheduling some time to help you with this.

jvanbaarsen:
@xadamy I can't really get this to work :( I need a bit more time to look into this issue.
@tombruijn I don't want this to be a holdback for the new redesign etc. For now I want to propose to go with the FORTAWESOME_TOKEN as env var.  Downside is that this means external people can't run this locally sadly. I do want to fix this, but I also want to take this blocker away for @jonasbrinkhoff.

jvanbaarsen:
@tombruijn For your idea: you can now set the fortawesome_token in the .npmrc, and don't have to think about it anymore.

jvanbaarsen:
@tombruijn Yeah same for me. If you have suggestions, please let me know.

PR #335, opened 2020-08-03
Update design to new layout

jonasbrinkhoff:
@jvanbaarsen and I decided to split up all changes to the documentation into separate PRs to make sure that we don't end up with a huge PR that will take ages to review and/or ship.

This PR updates the documentation to the new layout. It's a bit rough, will make some tweaks to it in the coming PRs. But I think it should be good enough for the moment :)


jvanbaarsen:
The font is terrible broken for me

source/assets/stylesheets/components/_grid.scss
```diff
@@ -0,0 +1,3 @@
+.grid-docs {
+  grid-template-rows: 4rem auto;
```
wesoudshoorn, line 2: Do we need this?

source/layouts/layout.html.haml
```diff
@@ -27,21 +27,25 @@
     %link{:rel => "stylesheet", :href => "https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css"}
     = partial "analytics"
 
-  %body
+  %body.font-rubik.bg-gray-100.pt-20
```
wesoudshoorn, line 30: If we make the header sticky, we don't need this pt-20?

source/layouts/layout.html.haml
```diff
@@ -27,21 +27,25 @@
...
+    .container.my-4.lg:my-12
+      .grid.grid-docs.grid-cols-1.lg:grid-cols-12
+        .row-start-1.row-span-1.lg:col-start-1.lg:col-span-3
+          .my-4.lg:pr-12
+            = partial "docsearch"
+        #navigation.row.start-3.lg:row-start-2.lg:row-span-1.lg:col-start-1.lg:col-span-3
+          .mt-4.lg:pr-12
+            = partial "side_nav"
+        #main.row-start-2.lg:row-start-1.lg:row-span-2.lg:col-start-4.lg:col-span-9
+          .bg-white.mt-8.lg:mt-0.px-4.lg:px-12.py-6.lg:py-8.rounded.shadow-sm
```
wesoudshoorn, line 44: Why the small shadow here? I feel like it could use a bit of a bigger shadow. It's such a big container.

wesoudshoorn: On medium screens, the horizontal padding feels a bit narrow. Maybe change it to this?

```suggestion
          .bg-white.mt-8.lg:mt-0.px-4.sm:px-8.lg:px-12.py-6.lg:py-8.rounded.shadow-sm
```
image

wesoudshoorn:
> The font is terrible broken for me

@jvanbaarsen can you elaborate on this? Screenshots? Description?

jvanbaarsen:
image
 The font just isn't there. Also no errors in console. So I guess it works for you because you have it installed on your machine.

source/assets/stylesheets/components/_grid.scss
jonasbrinkhoff: Yes unfortunately we do. I'll try to find a better solution for the future, but for now this is the easiest fix.

source/layouts/layout.html.haml
jonasbrinkhoff: Yes, header will be updated in a separate PR. I'll remove this it then. :)

PR #336, opened 2020-08-03
Add design language as non-submodule

jvanbaarsen:
Remove the design language submodule and add all the files as part of this repo.


source/assets/stylesheets/_design_language/_all.sass
```diff
@@ -0,0 +1,15 @@
+// Global imports
+@import "utilities/variables"
+@import "utilities/mixins"
+@import "utilities/placeholders"
+@import "utilities/reset"
```
jonasbrinkhoff, line 5:
```suggestion
```

Let's already remove the reset because it's really annoying and we don't need it (tailwind also has a basic reset)

jvanbaarsen: Removed it. Rebuilding now to see if everything still works.

PR #337, opened 2020-08-03
New header

jonasbrinkhoff:
Updates to the new header 🎉

The border are not working yet because of the CSS reset in the design language. This will be automatically fixed once https://github.com/appsignal/appsignal-docs/pull/336 is merged.


source/_header.html.erb
```diff
@@ -0,0 +1,122 @@
...
+      <div class="js-mobile-menu-toggle cursor-pointer lg:hidden toggle c-button c-button--sm c-button--white dark:c-button--gray">
+        Menu
+        <i class="fa fa-fw fa-bars ml-1"></i>
+      </div>
+
+      <nav class="js-mobile-menu-nav hidden lg:block flex-grow w-full lg:w-auto mt-6 lg:mt-0">
+        <ul class="lg:flex lg:justify-center">
+          <% navLinkClasses = "block lg:px-4 py-2 lg:dark:text-white-700 dark-hover:text-white border-t lg:border-t-0 lg:rounded no-underline" %>
+          <li>
+            <a href="/" class="<%= navLinkClasses %>">
```
wesoudshoorn, line 36: This should go to https://appsignal.com

source/_header.html.erb
```diff
@@ -0,0 +1,122 @@
...
+  ]
+%>
+
+<header class="sticky top-0 z-50 shadow">
+  <div
+    class="js-mobile-menu-wrapper bg-white dark:bg-gray-900 border-b border-gray-200 dark:border-gray-800 lg:relative"
+    data-mobile-menu-wrapper-active-classes="fixed inset-0"
+  >
+    <div class="container flex flex-wrap items-center py-5">
+      <a href="/" class="flex-grow lg:flex-grow-0 text-gray-800 dark:text-white h-6" title="Return to the homepage of AppSignal">
```
wesoudshoorn, line 23: This should go to https://appsignal.com

source/_header.html.erb
```diff
@@ -0,0 +1,122 @@
...
+                <li class="border-t py-2">
+                  <a href="https://appsignal.com/tour/<%= tour[:slug] %>" class="no-underline <%= tour[:text_class] %>">
+                    <%= tour[:title] %>
+                  </a>
+                </li>
+              <% end %>
+            </ul>
+          </li>
+          <li>
+            <a href="http://docs.appsignal.com" class="<%= navLinkClasses %>">Docs</a>
```
wesoudshoorn, line 92: Should we give this an active class?

source/_header.html.erb
jonasbrinkhoff: In the blog and in the current docs the logo always links back to the start of the blog/documentation. Should we change this behaviour?

I feel like that's kinda the problem with using one header for appsignal.com, blog.appsignal.com and docs.appsignal.com. 
- If you use the logo as escape hatch to the start documentation/blog, the behaviour of this button is different throughout the different pages. You could argue that it's consistent in that way, that it always leads back to start of the respective (sub-)domain, but not sure if that is very clear to the users.
- If you use the logo as escape hatch to the home appsignal.com the behaviour would be consistent, but users would have a hard time getting back to the start of the documentation or blog. Especially on mobile they'd have to open the menu and then click the respective link to get back to the start, which I imagine to be really annoying.

I'd still be interested to see how the behaviour of the users changes if we'd have different headers for the docs and blog. Otherwise, I think I prefer the way it's working right now.

source/_header.html.erb
jonasbrinkhoff: Yes, will do that!

wesoudshoorn:
Ah yes, we can keep the current behaviour for now. Logo goes to root of your current site, home goes to the homepage of appsignal.com

Our primary way of measuring how our blogposts now are successful is how many people click to the product tour. Using different headers breaks that flow. We can think about that in the future, but it's a bit too sticky to address that right now :-)

source/_header.html.erb
jvanbaarsen: I actually changed this behaviour on request of Milica on the blog today. The logo now goes to the same URL as the `home` button. Do I need to revert that?

jonasbrinkhoff:
@jvanbaarsen do you have the time to review this this week?

PR #340, opened 2020-08-04
Update doc search 

jonasbrinkhoff:
- Converts the doc search to the design system
- Slight design improvements


source/_docsearch.html.erb
```diff
@@ -0,0 +1,4 @@
+<div class="c-textfield">
```
wesoudshoorn, line 1: I'm still slightly annoyed by the gray border and the blue outline. I think that doing something with `border-blue-300` would be nice. We can ignore that for this specific PR though :-)

PR #341, opened 2020-08-04
Add "on this page" section

jonasbrinkhoff:
Adds an "on this page" section to the documentation

**To-Do**
- [x] Hide when empty
- [x] Improve font size
- ~[ ] Make it sticky~ Will be done in a separate PR

### DIFF
--- a/source/assets/javascripts/application.js
+++ b/source/assets/javascripts/application.js
@@ -3,6 +3,7 @@ import Cookies from "./js.cookie"
 import "./navigation_store"
 import "./dl_header"
 import { loadAnalyticsScripts } from "./analytics"
+import "./on_this_page.js"
 
 $(document).ready(function() {
   var navigationElement = $("#navigation");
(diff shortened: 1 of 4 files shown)


jvanbaarsen:
@jonasbrinkhoff I just pushed a thing that automates the generation of the "on this page"

wesoudshoorn:
@jonasbrinkhoff can you edit the description so people know what this does? It currently only shows the WIP description.

I'll review it now :-)

jonasbrinkhoff:
> Visually, I think we need to work on some alignment

Yes agreed. But it's quite hard to with all the moving parts in different PRs. Wanted to give it another design treatment once all major changes (sidenav, search bar, content and "on this page") are merged. 

> Making it sticky

Yes good idea, will do it!

> Two-line headers. Should we add a little separator between items to make this more clear? Or adjust the line-height/font-size/margins?

To me it seems quite clear. On a more general note, I think we should also try to keep headings short to make them easier to scan. So maybe that's a smarter way to solve the problem?

> Font-size

Yes, agreed. Will work on it.

> New header?

The PR is already done, just waiting for the review :)

jonasbrinkhoff:
Making it sticky requires some changes in the general layout of the page. Will do that in a separate PR.

PR #344, opened 2020-08-11
New sidenav

jonasbrinkhoff:
**New: Updated the behaviour of the side nav based on Toms suggestion**

Sub-sections and sub-pages now open freely while on a page, but then only the page you're currently in is open when you actually visit a page.

**Summary of changes:**
- Updated the design of the sidenav
- Separated the content and logic of the navigation. Navigation tree can now be changed in `/data/navigation.yml`
- Renamed a couple of files so that their active state is properly set by the `path_active` function in `config.rb` file


source/_side_nav.html.erb
```diff
@@ -0,0 +1,906 @@
+<%
+  sections = [
```
tombruijn, line 2: if we're going to define it like this, we should probably move it to a YAML file. Should be easier to edit.

config.rb
```diff
@@ -77,6 +77,14 @@ def site_name
     "AppSignal documentation"
   end
 
+  def path_active?(path)
+    if path === '/'
+      return current_page.url === '/'
+    end
+
+    return current_page.url.include?(path)
```
tombruijn, line 85:
```suggestion
    current_page.url.include?(path)
```
Unnecessary return

source/_side_nav.html.erb
```diff
@@ -0,0 +1,906 @@
...
+  ]
+
+%>
+
+
+
+
+<nav>
+  <ul>
+    <% sections.each do |section| %>
```
tombruijn, line 857: This whole thing seems really complex.

I don't like how I need to first click a section to wait for a page load and then only see the section links. That means it takes me X amount of clicks with page loads to go to the page I need to visit. Makes the navigation slower.

I'd rather keep the open and close logic in JavaScript.

jonasbrinkhoff: I talked with Wes about this (and I even implemented the old navigation style), but we came to the conclusion to keep it like it is now.

With the old navigation style, you either end up with a really long navigation (because multiple sub-navigations are open) or you'd loose the context of where you currently are (because once you opened another sub-navigation, the currently active one would close).

We are aware that the new solution increases the amount of clicks and time to navigate through the page, but we think that the drawbacks of the other solutions are bigger. In the long term, we'd like to move the documentation to Gatsby (or something similar) which would drastically reduce page load times and thereby hopefully eliminate the problems with the new navigation.

source/_side_nav.html.erb
jonasbrinkhoff: Done :)

source/_side_nav.html.erb
tombruijn: > With the old navigation style, you either end up with a really long navigation (because multiple sub-navigations are open) or you'd loose the context of where you currently are (because once you opened another sub-navigation, the currently active one would close).

There are other solutions to that problem as discussed. Another idea is to freely open sections while on a page, but then only the section you're in is open when you actually visit a page.

> In the long term, we'd like to move the documentation to Gatsby (or something similar) which would drastically reduce page load times and thereby hopefully eliminate the problems with the new navigation.

Load times are not the issue. As a user, on the deployed site, you're still waiting for page load times. That's my problem.

jonasbrinkhoff:
@wesoudshoorn 

> The links are now rendered directly inside an, which is unvalid HTML

I think something is missing here in this comment?

> We used to have a single side navigation and hid subitems with JS, we now have everything hidden based on Ruby

Updated the code, non-active sub-navigation elements are now rendered but hidden.

> Metrics and standalone agent links do not work

Fixed all links now

jonasbrinkhoff:
Updated all page names & links now, all links should now work properly in the navigation.

Also refactored the navigation a bit to make it more readable.

wesoudshoorn:
@jonasbrinkhoff https://deploy-preview-344--dreamy-wright-e9a95c.netlify.app/ the deploy preview is broken now?

data/navigation.yml
```diff
@@ -0,0 +1,631 @@
...
+                link: "/front-end/plugins/plugin-path-decorator.html"
+          -
+            name: "Troubleshooting"
+            link: "/front-end/troubleshooting.html"
+
+  -
+    name: "Features"
+    sub_sections:
+      -
+        name: "Standalone agent<sup>Beta</sup>"
```
wesoudshoorn, line 508:
```suggestion
        name: "Standalone agent <sup>Beta</sup>"
```

source/_side_nav.html.erb
jonasbrinkhoff: > Another idea is to freely open sections while on a page, but then only the section you're in is open when you actually visit a page.

Really like this idea, implement it :)

jonasbrinkhoff:
Updated the behaviour of the side nav based on Toms suggestion and moved around some pages to make it work. Would be great if you could review it again :)

jvanbaarsen (approved):

jvanbaarsen:
@jonasbrinkhoff What is the status of this? Only waiting for reviews?

jonasbrinkhoff:
> @jonasbrinkhoff What is the status of this? Only waiting for reviews?

Yes, only waiting for reviews!

data/navigation.yml
```diff
@@ -0,0 +1,697 @@
...
+          -
+            name: "Settings"
+            link: "/application/settings/"
+            sub_pages:
+              -
+                name: "Application settings"
+                link: "/application/settings/application-settings.html"
+              -
+                name: "Notification settings"
+                link: "/application/settings/notification-settings.html"
```
tombruijn, line 55: I wouldn't want to move pages in this PR. Yes it's now matching the new "navigation structure", but we're going to restructure it later in the docs improvement project. We then need to move it again.

The problem being that moving files require redirects to not break existing links. (This PR doesn't update those redirects as defined in `netlify.toml`.) So in this PR we would need to add more redirects, then later in the docs improvements projects we need to add even more redirects. And by then we've exhausted netlify's redirection list, if they still only support a limited amount of redirects.

tombruijn (changes requested):

config.rb
```diff
@@ -77,6 +77,14 @@ def site_name
     "AppSignal documentation"
   end
 
+  def path_active?(path)
+    if path === '/'
+      return current_page.url === '/'
```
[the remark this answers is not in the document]
jonasbrinkhoff: `current_page.url.include?("/")` would be true for all paths. That's why, if the given `path` is `"/"`, it should only return true if the `current_path` is also `"/"`.

config.rb
```diff
@@ -77,6 +77,14 @@ def site_name
...
+    end
+
+    current_page.url.include?(path)
```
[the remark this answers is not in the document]
jonasbrinkhoff: Oh, good point! Didn't knew `start_with?` exists. I'll update it :)

data/navigation.yml
```diff
@@ -0,0 +1,697 @@
...
+        icon: "fas fa-question-circle"
+        link: "/faq.html"
+      -
+        name: "Your applications"
+        icon: "fas fa-server"
+        link: "/application/"
+        pages:
+          -
+            name: "Overview"
+            link: "/application/overview.html"
```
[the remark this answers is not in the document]
jonasbrinkhoff: I generally agree, but this would break the `path_active` function because of how `current_page.url` works.

Example:
- You rename `/application/overview.html` file to `/application/index.html`
- You then visit https://docs.appsignal.com/application/index.html
- You'd expect `current_page.url` to return `/application/index.html`, but in reality it returns just `/application/`
- And since `"/application/".start_with?("/application/index.html")` is `false`, `path_active` would return `false`

If you have any idea how to fix this, let me know.

config.rb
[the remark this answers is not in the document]
tombruijn: Yes, but that would also be covered under `path == current_path.url`. That should then also work for most paths that we can match 1 to 1 and don't have to check for a partial match with `include?`/`start_with?`.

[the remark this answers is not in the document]
jonasbrinkhoff: I'm not sure if I got your comment. But we need a partial check, to determine the active state of parent elements.

Here's how `path_active` should work like, maybe this makes things more clear:
- when on `"/"`
  - `path_active("/")` should be `true`
  - `path_active("/application/")` should be `false`
- when on `"/application/"`
  - `path_active("/")` should be `false` <- weird case
  - `path_active("/application/")` should be `true`
  - `path_active("/application/index.html")` should be `true` <- weird case (only exists if we use index.html)
  - `path_active("/application/new-application.html")` should be `false`

tombruijn:
@jonasbrinkhoff I've renamed all the overview.html to index.html and updated the logic in the last commit. Can you let me know if it works good now?

tombruijn:
I'll have a look what pages should be renamed back to their originals. I don't think the pages under "Your application > Settings" should really go there. I've put them there to give some form of structure, but every page upon themself is a feature  we should put in a better structure in the future.

tombruijn:
@jonasbrinkhoff I've redirected and reverted some files, see last commit. I think it's fine to merge now.

jonasbrinkhoff:
> @jonasbrinkhoff I've redirected and reverted some files, see last commit. I think it's fine to merge now.

Amazing, thanks for your help!

PR #347, opened 2020-08-18
Turn sidenav sub-sections and sub-pages into accordions

jonasbrinkhoff:
> Another idea is to freely open sections while on a page, but then only the section you're in is open when you actually visit a page.

Created a quick prototype based on this idea. There are still some minor UX flaws, but I think I prefer this solution over the current one.

[skip review]


wesoudshoorn:
@jonasbrinkhoff I like this a lot, probably a good balance between all the options:

- Clean sidenavigation
- Go as deep as you need right away

The only thing I'd consider is consistent hinting of sub-items. The start + FAQ page are the only ones without subnavigation, but you have to hover to figure that out. I'd also be fine with moving the navigation around a bit so there's no item without subnavigations left?

jonasbrinkhoff:
> The only thing I'd consider is consistent hinting of sub-items

Yes, agreed. But I feel like that this is not a dealbreaker, so for now we could just merge it as it is and fix this issue later? (So we can finally wrap up this side project 🙃)

PR #366, opened 2020-10-01
Add the right key required for javascript package

shairyar:
On the front-end when `@appsignal/javascript` package is being used, the key which is required is Front-end error monitoring key and not the Push Key.


xadamy (approved):

xadamy:
Thanks @shairyar! FWIW I also checked the READMEs but the wording is good there. 👍

shairyar:
Thanks @xadamy

PR #368, opened 2020-10-05
add example usage of tracer and span

shairyar:
This PR adds instructions on the usage of tracer and span.

### DIFF
--- a/source/nodejs/tracing/span.html.md
+++ b/source/nodejs/tracing/span.html.md
@@ -111,6 +111,7 @@ span.setSampleData("custom_data", { foo: "bar" });
 This key accepts nested objects and will result in the following block on a Incident Sample page for both Exception and Performance samples formatted as JSON.
 
 ![custom_data](/assets/images/screenshots/sample_data/custom_data.png)
+
 ## Example implementation
 Here is an example implementation of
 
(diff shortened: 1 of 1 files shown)


source/nodejs/tracing/span.html.md
```diff
@@ -111,3 +111,23 @@ span.setSampleData("custom_data", { foo: "bar" });
...
+- Closing the child span.
+
+```js
+  const tracer = appsignal.tracer();
+  const span = tracer.currentSpan();
+
+  const childSpan = span.child();
+  childSpan.setName("GET /homepage");
+
+  span.setSampleData("custom_data", { foo: "bar" });
```
matsimitsu, line 130: I'd remove this line, it might be confusing to set things on the span, while the subject is the child span. 

mabe add a comment `#do stuff` to indicate that the `childSpan.close()` wraps the code inside.

matsimitsu (changes requested):

source/nodejs/tracing/span.html.md
shairyar: yes makes sense.

source/nodejs/tracing/span.html.md
```diff
@@ -111,3 +111,22 @@ span.setSampleData("custom_data", { foo: "bar" });
 This key accepts nested objects and will result in the following block on a Incident Sample page for both Exception and Performance samples formatted as JSON.
 
 ![custom_data](/assets/images/screenshots/sample_data/custom_data.png)
```
matsimitsu, line 113: Needs a newline here to not break the link above, can be merged after newline is added and build is green.

matsimitsu (approved):

PR #375, opened 2020-10-20
Add guides section with deploy markers guide

tombruijn:
Add first guide about deploy markers to the new guides section in the
docs. This section will be more hands on in taking people through the
steps we recommend they take to get the most out of AppSignal.

I started with the deploy markers guide because that would be my first
recommendation after an app reports data to AppSignal.

There's a list of links to other guides that have yet to be written.

### DIFF
--- a/source/guides/deploy-markers.html.md
+++ b/source/guides/deploy-markers.html.md
@@ -92,7 +92,13 @@ Are deploys not being reported or incorrectly? [Contact us][contact] and we will
 
 ## Next steps
 
+- [Grouping parts of your app with namespaces](/guides/namespaces.html) - next guide
 - [Setting up backtrace links][error backtrace links]
+
+---
+
+- [Add a new application](/guides/new-application.html) - previous guide
+- [Getting started guides](/guides/) - Guides overview
 
 [deploy markers]: /application/markers/deploy-markers.html
 [heroku support]: /application/markers/deploy-markers.html#heroku-support
--- a/source/guides/deploy-markers.html.md
+++ b/source/guides/deploy-markers.html.md
@@ -1,0 +1,101 @@
+---
+title: "Reporting deploys to track improvements"
+---
+
+Every time an app gets deployed, changes that affect the app start running. By tracking deploys in AppSignal, error incidents and performance measurements are grouped per deploy. They will also allow for AppSignal.com to link directly from an [error backtrace]([error backtrace links]) to the line of code in your app for that version of the app.
+
+We will track deploys in AppSignal using deploy markers. Read more about [deploy markers], what they're used for, how to send them, and more.
+
+## Configuring AppSignal
+
+To report deploys to AppSignal we will use [deploy markers]. In this guide we will focus on the `revision` config option method of reporting deploy markers. (Other methods may be available, more details about why we prefer this method can be found on the [deploy markers] page.)
+
+The `revision` config option is configured differently per integration language. See the list of integrations below for the one your app uses.
+
+Heroku apps using the "Dyno Metadata" labs feature will [automatically report deploys][heroku support].
+
+### Automatically update the revision
+
+We recommend fetching the revision config option value from the SCM (Git/SVN/other) system dynamically on start or on deploy. This way the value doesn't have to be manually updated per deploy. If a Git SHA (or other SCM revision value) is used it will also be compatible with [error backtrace links].
+
+An example for Git:
+
+```bash
+git log --pretty=format:'%h' -n 1
+# Will output the current commit's short SHA
+```
+
+We will use the Git example in the integration examples below.
+
+### Ruby
+
+In the Ruby integration we will load the Git revision in the `config/appsignal.yml` config file. By using ERB we call the `git log` command and set the output as the `revision` config option.
+
+```ruby
+# config/appsignal.yml
+production:
+  revision: "<%= `git log --pretty=format:'%h' -n 1` %>"
+  # Other config
+```
+[Ruby `revision` configuration option details](/ruby/configuration/options.html#option-revision)
+
+### Elixir
+
+In the Elixir integration we will load the Git revision in the `config/appsignal.exs` config file (your file location may differ). In this `.exs` command we call the `git log` command with Elixir and set the output as the `revision` config option.
+
+```elixir
+# config/appsignal.exs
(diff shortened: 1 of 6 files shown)


source/guides/deploy-markers.html.md
```diff
@@ -0,0 +1,82 @@
+---
+title: "Deploy markers"
```
tombruijn, line 2: @shairyar what do you think of this as a first guide?

source/guides/deploy-markers.html.md
```diff
@@ -0,0 +1,82 @@
...
+config :appsignal, :config,
+  revision: revision
+  # Other config
+```
+[Elixir `revision` configuration option details](/elixir/configuration/options.html#option-revision)
+
+### Node.js
+```javascript
+const appsignal = new Appsignal({
+  revision: "REVISION PLACEHOLDER"
```
tombruijn, line 52: I'd need some help with this one and the JavaScript one to automatically fetch the git SHA here.

source/guides/deploy-markers.html.md
```diff
@@ -0,0 +1,82 @@
...
+# Will output the current commit's short SHA
+```
+
+We will use the Git example in the integration examples below.
+
+### Ruby
+```ruby
+# config/appsignal.yml
+production:
+  revision: "<%= `git log --pretty=format:'%h' -n 1` %>"
```
tombruijn, line 34: We can also opt to load from an system environment variable, but that could also become more confusing. WDYT?

source/guides/deploy-markers.html.md
shairyar: It is a good start, from a guide point of view this is what i was thinking.

1. Creating an account
2. Creating an organisation
3. Creating an application
4. Integrating AppSignal
5. rest of the features of appsignal
6. ...

When I started reading the guide it felt more like the about section we already have for deploy markers, should we include more screenshots? like where this deploy can be accessed on dashboard, how it helps in filtering errors on errors page? where on graph we can see the marker, details like this.

source/guides/deploy-markers.html.md
shairyar: We could add that as a tip or info box for the users to know that environment variables can be used as well. Yesterday I was playing around with setting up AppSignal with one of app i have on DigitalOcean, the deployment service i use is hatchbox.io, this service creates a REVISION file in the root of the app every time there is a new deploy so adding instructions on how to read the revision from a file will also be nice.

source/guides/deploy-markers.html.md
tombruijn: I'm skipping over the starter topics for now, like user account, organization and app, and diving into the areas that are probably more interesting for users that have already set up AppSignal. Because the user account creation, organization and app are already part of the onboarding wizard on AppSignal.com.

> When I started reading the guide it felt more like the about section we already have for deploy markers

I'm repeating some content yes, introducing the concept of a deploy marker and why they are useful, but the deploy markers page has way more content about this: https://docs.appsignal.com/application/markers/deploy-markers.html

> should we include more screenshots? like where this deploy can be accessed on dashboard, how it helps in filtering errors on errors page? where on graph we can see the marker, details like this.

Yes, screenshots or a video will of course be a lot of help in making these guides a lot easier to understand.

But is the guide good so far in terms of content?

source/guides/deploy-markers.html.md
tombruijn: Yes we can add another guide for example that just talks about how to leverage a system to read the app revision. I think adding that to this guide can make it very long.

For that reason I'm also not explaining in detail how the AppSignal config works, I think a guide about just configuration would be useful too as a separate guide.

source/guides/deploy-markers.html.md
shairyar: Sounds good to me and yes in terms on content it seems good to me.

source/guides/deploy-markers.html.md
shairyar: ok

source/guides/deploy-markers.html.md
tombruijn: @jeffkreeftmeijer do you know how to call a process in Node.js to get the current git commit SHA?

source/guides/deploy-markers.html.md
```diff
@@ -0,0 +1,81 @@
...
+  revision: "REVISION PLACEHOLDER"
+  // Other config
+})
+```
+[Node.js `revision` configuration option details](/nodejs/configuration/options.html#option-revision)
+
+### JavaScript
+```javascript
+const appsignal = new Appsignal({
+  revision: "REVISION PLACEHOLDER"
```
tombruijn, line 61: I don't think we can call the `git log` command in JavaScript (as it's running in the browser) so we'd need to set it during production compilation I guess? Any ideas? What build tools to use as an example, etc.

source/guides/deploy-markers.html.md
tombruijn: Ah found _a way_ I think:

```diff
+var childProcess = require("child_process");
+var REVISION = childProcess.execSync("git log --pretty=format:'%h' -n 1").toString();
 const appsignal = new Appsignal({
-  revision: "REVISION PLACEHOLDER"
+  revision: REVISION
```

I'll use that for now

source/guides/deploy-markers.html.md
```diff
@@ -0,0 +1,81 @@
...
+production:
+  revision: "<%= `git log --pretty=format:'%h' -n 1` %>"
+  # Other config
+```
+[Ruby `revision` configuration option details](/ruby/configuration/options.html#option-revision)
+
+### Elixir
+```elixir
+# config/appsignal.exs
+{revision, _exitcode} = System.cmd("git", ["log", "--pretty=format:%h", "-n 1"])
```
tombruijn, line 42: @jeffkreeftmeijer can you let me know if this is any good as a method to fetch the git sha in Elixir?

source/guides/deploy-markers.html.md
shairyar: @xadamy/ @thijsc  might be the best people to answer this. I myself am not sure.

tombruijn:
@shairyar I've updated the page with instructions for every integration. It may not be the perfect solution (as I don't know the specifics or recommended methods for Elixir, Node.js and JavaScript), but at least there's something. Any improvements can be made later.

shairyar (approved):
Looks good to me

PR #386, opened 2020-10-26
Add session data filtering guide

tombruijn:
Merge the Ruby and Elixir session data filtering pages into one guide.
I couldn't find any information about this "feature" in the JavaScript
or Node.js sections, and couldn't really tell if it's entirely
implemented based on their code bases, so I omitted these two
integrations for now. Feel free to add them if you are sure how they
work!

I added a new page that explains the session data filtering in some more
detail.

I've moved the "filter out all session data" part of the Ruby and Elixir
pages to the /application/session-data-filtering.html page because it
didn't feel like the best place for it was the guide. The "feature" page
seems like a better fit, for when people are reading further up on it.


shairyar (approved):

xadamy:
It works the same in Node.js, but you pass `filterSessionData` to the `Appsignal` constructor. You can find these options by looking in `configmap.ts` in the repo.

tombruijn:
Thanks @xadamy! I've updated the page. Merging.

Issue #371, opened 2020-10-14
Fix config options for Node.js in docs


xadamy:
This missed the end of the last cycle as I wanted to get the rest of the docs out for support to use, now this needs to be fixed. Alternatively, it could be rolled into a larger project where the options YAML file is split into separate files so it's more manageable (as suggested by @tombruijn in #308).

tombruijn:
https://github.com/appsignal/appsignal-docs/pull/387

PR #387, opened 2020-10-26
Nodejs config options

thijsc:
I've added the Node.js config options I know to work with basic descriptions. Leaving #371 open since this still needs more attention.

I could not get Markdown rendering to work and I do not understand why. Got this error:

```NoMethodError at /nodejs/configuration/options.html
undefined method `link_to' for nil:NilClass

Ruby	/Users/thijs/.rbenv/versions/2.4.5/lib/ruby/gems/2.4.0/gems/middleman-core-4.2.1/lib/middleman-core/renderers/redcarpet.rb: in link, line 89
Web	GET localhost/nodejs/configuration/options.html
```

I've added a temp fix so I could see the page in 79b3b2f. @tombruijn do you maybe have any idea what could be causing this error? I've looked quite hard but could not find a reason.

### DIFF
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -122,18 +122,18 @@ options:
       default_value: null
       description: |
         Name of your application as it should be displayed on AppSignal.com. If you use Ruby on Rails the gem will auto-detect the name and you can leave this empty. For other frameworks setting this is mandatory.
-
-        -> **Note**: Changing the name or [environment](#option-appsignal_app_env) of an existing app will create a new app on AppSignal.com.
     elixir:
       type: atom
       default_value: null
+      description: |
+        Name of your application as it should be displayed on AppSignal.com.
     nodejs:
       config_key: appName
       type: string
       default_value: nil
+      description: |
+        Name of your application as it should be displayed on AppSignal.com.
     description: |
-      Name of your application as it should be displayed on AppSignal.com.
-
       -> **Note**: Changing the name or [environment](#option-env) of an existing app will create a new app on AppSignal.com.
   - config_key: push_api_key
     env_key: APPSIGNAL_PUSH_API_KEY
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -108,7 +108,7 @@ options:
 
         -> **Note**: Changing the [name](#option-name) or environment of an existing app will create a new app on AppSignal.com.
     nodejs:
-      config_key: appEnv
+      config_key: environment
       type: string
       default: env
       description: |
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -146,7 +146,7 @@ options:
       type: string
       default_value: null
     nodejs:
-      config_key: pushApiKey
+      config_key: apiKey
       type: string
       default_value: null
     description: |
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -128,7 +128,6 @@ options:
       description: |
         Name of your application as it should be displayed on AppSignal.com.
     nodejs:
-      config_key: appName
       type: string
       default_value: nil
       description: |
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -17,6 +17,10 @@ options:
       type: bool
       default_value:
         markdown: "`false` / detected by system"
+    nodejs:
+      type: bool
+      default_value:
+        markdown: "`false` / detected by system"
     description: |
       Configure AppSignal to be active or not for a given environment. Most commonly used in the [file configuration](index.html) per environment.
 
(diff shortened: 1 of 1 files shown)


config.rb
```diff
@@ -126,7 +126,7 @@ def edit_link
 
   def markdown(content)
     # https://github.com/hashicorp/middleman-hashicorp/blob/14c705614b2f97b5a78903f17b904f767c1fdbe2/lib/middleman-hashicorp/redcarpet.rb
-    Redcarpet::Markdown.new(AppsignalMarkdown, AppsignalMarkdown::OPTIONS).render(content)
+    Redcarpet::Markdown.new(Redcarpet::Render::HTML, AppsignalMarkdown::OPTIONS).render(content)
```
tombruijn, line 129: Please undo this change so the markdown works the same everywhere. Now the special markdown syntax doesn't work on the config options page.

data/config_options.yml
```diff
@@ -192,54 +214,37 @@ options:
...
-        If you're affected by our [DNS timeouts](/support/known-issues.html#dns-timeouts), try setting a DNS server manually using this option that doesn't use more than __4__ dots in the server name.
+      If you're affected by our [DNS timeouts](/support/known-issues.html#dns-timeouts), try setting a DNS server manually using this option that doesn't use more than __4__ dots in the server name.
 
-        - Acceptable values: `8.8.8.8`, `my.custom.local.server`.
-        - Not acceptable values: `foo`, `my.awesome.custom.local.dns.server`.
+      - Acceptable values: `8.8.8.8`, `my.custom.local.server`.
+      - Not acceptable values: `foo`, `my.awesome.custom.local.dns.server`.
 
-        If the DNS server cannot be reached the agent will fall back on the host's DNS configuration and output a message in the `appsignal.log` file: `A problem occurred while setting DNS servers`.
+      If the DNS server cannot be reached the agent will fall back on the host's DNS configuration and output a message in the `appsignal.log` file: `A problem occurred while setting DNS servers`.
```
tombruijn, line 247: You've removed the Ruby description above and indented this description on to the left so it applies to all languages, but only has Elixir examples. Please add a description per language as before.

tombruijn (changes requested):

tombruijn:
> I've added a temp fix so I could see the page in 79b3b2f. @tombruijn do you maybe have any idea what could be causing this error? I've looked quite hard but could not find a reason.

From what I can tell you, after middleman starts, you first need to visit a page that's not the config page, then visit the config page and it renders correctly.

config.rb
thijsc: I mentioned this as being a temp fix, I had no intention to merge this.

thijsc:
> From what I can tell you, after middleman starts, you first need to visit a page that's not the config page, then visit the config page and it renders correctly.

Ah so this is a known issue?

tombruijn:
> Ah so this is a known issue?

Not that I know. I just noticed the other day.

data/config_options.yml
```diff
@@ -132,6 +144,9 @@ options:
       since: 0.1.0
       type: string
       default_value: null
+    nodejs:
```
tombruijn, line 147: All the Node.js config options need to have a camelcase version instead of the underscored versions. You can add a `config_key` override for every Node.js config option YAML block.

```suggestion
    nodejs:
      config_key: "pushApiKey"
```

thijsc:
Processed feedback and removed the temporary fix.

data/config_options.yml
```diff
@@ -117,10 +127,14 @@ options:
     elixir:
       type: atom
       default_value: null
-      description: |
-        Name of your application as it should be displayed on AppSignal.com.
+    nodejs:
+      config_key: appName
```
tombruijn, line 131:
```suggestion
```

It's the same as other integrations with `name`:
https://github.com/appsignal/appsignal-nodejs/blob/a75027dacc3ba1920e579218350f64c09a606d1b/packages/nodejs/src/config/configmap.ts#L4

data/config_options.yml
```diff
@@ -103,6 +107,12 @@ options:
         ```
 
         -> **Note**: Changing the [name](#option-name) or environment of an existing app will create a new app on AppSignal.com.
+    nodejs:
+      config_key: appEnv
```
tombruijn, line 111:
```suggestion
      config_key: environment
```

Should be `environment` according to this map: https://github.com/appsignal/appsignal-nodejs/blob/a75027dacc3ba1920e579218350f64c09a606d1b/packages/nodejs/src/config/configmap.ts#L5

data/config_options.yml
```diff
@@ -132,6 +146,10 @@ options:
       since: 0.1.0
       type: string
       default_value: null
+    nodejs:
+      config_key: pushApiKey
```
tombruijn, line 150: My bad, this is actually `apiKey` according to this: https://github.com/appsignal/appsignal-nodejs/blob/a75027dacc3ba1920e579218350f64c09a606d1b/packages/nodejs/src/config/configmap.ts#L3

```suggestion
      config_key: apiKey
```

tombruijn (commented):
Fixed some stuff with some descriptions in an added commit..

Can you double check my suggestions for the key names?

thijsc:
Thanks for checking @tombruijn

PR #396, opened 2020-10-28
Add configuration guide

tombruijn:
Add guide to help users configure AppSignal using an example config
option: debug/namespace.

This guide can be useful as first introduction to configuration in
AppSignal, but can also be linked from other parts of the docs and
guides to help users set any config options we mention in docs and
examples.

### DIFF
--- a/source/guides/index.html.md
+++ b/source/guides/index.html.md
@@ -7,6 +7,7 @@ You have set up AppSignal and data is coming in from your app. What's next?
 In the next couple steps we'll guide you through ways to make the data being reported from your app more useful. Follow the guides below in a step-by-step process, or jump ahead to the guide that seems interesting for your app.
 
 1. [Add a new application](/guides/new-application.html)
+1. [Configuring applications](/guides/configuration.html)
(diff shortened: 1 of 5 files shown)


source/guides/index.html.md
```diff
@@ -7,6 +7,7 @@ You have set up AppSignal and data is coming in from your app. What's next?
 In the next couple steps we'll guide you through ways to make the data being reported from your app more useful. Follow the guides below in a step-by-step process, or jump ahead to the guide that seems interesting for your app.
 
 1. [Add a new application](/guides/new-application.html)
+1. [Configuring applications](/guides/configuration.html)
```
tombruijn, line 10: What do you think, list as one of the earlier guides? Or only link to this guide from other guides to help explain and guide users through setting a config option.

For example, in this guide about reporting deploy markers by setting the `revision` config option: https://docs.appsignal.com/guides/deploy-markers.html#ruby
We can link to the config guide from there, any other similar location where the docs talk about config options. Maybe we should even add an information icon like this one: ℹ️ to link to this guide rather than have a text based link for every such occurrence.

shairyar (approved):

PR #406, opened 2020-11-26
fix header navi links to /ruby and homepage

aksie:


source/_header.html.erb
```diff
@@ -55,7 +55,7 @@
     data-mobile-menu-wrapper-active-classes="fixed inset-0 overflow-y-auto"
   >
     <div class="container flex flex-wrap items-center py-5 px-8">
-      <a href="/" class="flex-grow lg:flex-grow-0" title="Return to the homepage of AppSignal">
+      <a href="https://appsignal.com" class="flex-grow lg:flex-grow-0" title="Return to the homepage of AppSignal">
```
tombruijn, line 58: This original change should be good, as it links to the root of the domain. With a `/` it also works when we're running the app in development mode. I would revert this change.
```suggestion
      <a href="/" class="flex-grow lg:flex-grow-0" title="Return to the homepage of AppSignal">
```
Are you seeing an issue with the `/` path?

aksie: @tombruijn 
Currently the logo links to the root of the (sub)domain docs.appsignal.com but it should link to the root of appsignal.com, hence the proposed change. I hope that explains it better?

tombruijn: Ah yes! That's more clear for me :)

tombruijn (approved):

PR #408, opened 2020-12-01
Add missing backtick to upgrade guide

cybrox:
Minor inconvenience I noticed while reading through the guide.
The missing backtick causes the text to be highlighted instead of the inline code.


tombruijn (approved):
Good catch! Thanks for the fix!

PR #412, opened 2020-12-18
Restore the Elixir 1.x documentation

jeffkreeftmeijer:
Restore the Elixir 1.x documentation for users that need to downgrade. The
documentation pages can be accessed by adding `1.x` to the URL, like so:

https://docs.appsignal.com/elixir/installation/ becomes
https://docs.appsignal.com/elixir/1.x/installation/.

The 1.x docs aren’t linked in the sidebar to avoid confusion. This might not be the nicest way to go about it, but this at least gives us a way to link to the docs when asking users to downgrade.


tombruijn (approved):
Interesting problem. We probably also need this for Ruby if it changes to the span API. Let's see how often we need to link to these docs.

thijsc (approved):

thijsc:
Pragmatic solution for now, will indeed require more attention later for Ruby. cc @shairyar, relevant to know if somebody asks for the old docs in support.

shairyar:
Noted

PR #413, opened 2021-01-06
Add note related to uploading multiple sourcemap files

shairyar:
Just adding a note related to uploading of multiple sourcemap files.


xadamy:
Just a note to say that the Webpack plugin can now upload multiple sourcemaps as of 1.1.0!

matsimitsu (approved):

PR #420, opened 2021-01-20
Remove Ruby's garbage collection config option

shairyar:
Removing Ruby's garbage collection config option since the feature still requires some work.


jvanbaarsen:
@shairyar Is this something that came out of a discussion within the Integrations team? I don't know enough about the integration to be able to say "Yes this needs more work" I'll add @tombruijn to this; He knows more about this.

shairyar:
@thijsc and I were discussing this in the [slack](https://appsignal.slack.com/archives/CNPP953E2/p1611125226001100) today

thijsc (approved):

tombruijn (approved):
Should we also remove it from the gem itself, or just hide it from the docs like this?

shairyar:
I am guessing if it is only the dashboard changes that are pending we can leave it in otherwise remove it?

PR #422, opened 2021-01-26
Make Elixir log_path doc consistent with reality

thijsc:
The log path should contain just a path. The integration will add the `appsignal.log` suffix to it.

### DIFF
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -415,10 +415,10 @@ options:
       default_value: ./log
     elixir:
       type: string
-      default_value: "/tmp/appsignal.log"
+      default_value: "/tmp"
       since: 1.0.0
       since_notes:
-        - "`1.9.0`: The default log path was changed from `/tmp/appsignal/appsignal.log` to `/tmp/appsignal.log`."
+        - "`1.9.0`: The default log path was changed from `/tmp/appsignal/` to `/tmp/`."
     nodejs:
       config_key: logPath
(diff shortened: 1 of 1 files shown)


data/config_options.yml
```diff
@@ -415,10 +415,8 @@ options:
       default_value: ./log
     elixir:
       type: string
-      default_value: "/tmp/appsignal.log"
+      default_value: "/tmp"
       since: 1.0.0
-      since_notes:
-        - "`1.9.0`: The default log path was changed from `/tmp/appsignal/appsignal.log` to `/tmp/appsignal.log`."
```
tombruijn, line 421: The note can remain, but the file name should be removed right?

tombruijn (approved):

jeffkreeftmeijer (approved):
Approved, aside from the `since_notes` issue Tom raised.

data/config_options.yml
thijsc: I think it's not actually correct? It would actually be `/tmp/appsignal` as a value then.

tombruijn: I can't make a suggestion on a deleted line but I was thinking this as the updated note. The note was added there for a reason 🤷 

>          - "`1.9.0`: The default log path was changed from `/tmp/appsignal/` to `/tmp/`."

thijsc: added

PR #424, opened 2021-02-01
Fix broken Web Vitals Reporting example app

bartdeslagmulder:
* createServer wasn't imported
* PORT wasn't used on .listen(PORT) and in the console.log


tombruijn (approved):
Great catch @bartdeslagmulder ! Thanks for the PR!

PR #425, opened 2021-02-03
Update appsignal absinthe plug to 2.x

mateusmaso:
After version `2.x.x` the documented solution stopped working. Instead of simply replacing `Transaction.set_action/1` to `Appsignal.Plug.put_name/2` it is also important to return the new `Plug.Conn` struct (containing private attribute `:appsignal_name`) inside plug `call/2` otherwise it won't work.

### DIFF
--- a/source/elixir/integrations/absinthe.html.md
+++ b/source/elixir/integrations/absinthe.html.md
@@ -4,7 +4,7 @@ title: "Absinthe <sup>Unofficial support</sup>"
 
 !> **Warning:** Use this solution at your own risk.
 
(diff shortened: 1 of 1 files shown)


tombruijn:
Hi @mateusmaso, Thanks for the PR!

@shairyar have you been testing this change already?

shairyar:
@tombruijn not yet, it came in last night it seems so I will taking a look at this today.

tombruijn (approved):

shairyar (approved):

PR #428, opened 2021-02-16
Mention the right package to add

shairyar:
Replacing the name of the package to mention the correct package to add.


tombruijn (changes requested):
From the npm page of the pg package: https://www.npmjs.com/package/@appsignal/pg

> This package has been deprecated. pg is now instrumented automatically by @appsignal/nodejs, meaning that this package is no longer necessary. Please remove this module from your package.json as it is now a no-op.

So I think the page is correct.

shairyar:
Oh, I overlooked that, thank you for pointing that out Tom

PR #445, opened 2021-04-19
Update incorrect Cloud66 documentation

LimeBlast:
The suggested code doesn't work for cloud 66, as the environment variable in the example doesn't seem to exist. I also believe having it as a root-level item in the yaml file doesn't work.

I've replaced it with code that we use within our stack.


shairyar:
Thank you for sharing that, if the configuration for `revision` on cloud66 is the same as defined under ruby [docs](https://docs.appsignal.com/guides/deploy-markers.html#ruby) then there is no point in having the same instructions in 2 different places so I am going to close it.

https://github.com/appsignal/appsignal-docs/pull/448

PR #447, opened 2021-04-22
Updating the uptime monitoring image

shairyar:
Now that we introduced the check for HTML body content so the image needs to be updated.


tombruijn (approved):

tombruijn:
Can you run imageoptim over the file and amend it to this commit? So it's a smaller file size?

PR #454, opened 2021-05-05
Mention the order of Appsignal.Phoenix and Plug.Builder

shairyar:
`Plug.Builder` redefines the call/2 function we’ve overwritten with the use `Appsignal.Phoenix` call so we need to call`Plug.Builder` before `Appsignal.Phoenix` call.

### DIFF
--- a/source/elixir/installation/umbrella.html.md
+++ b/source/elixir/installation/umbrella.html.md
@@ -30,7 +30,7 @@ application](/elixir/integrations/phoenix.html). One thing to keep in mind is
 to `use Appsignal.Phoenix` in your application’s endpoint file. If you have
 multiple nested Phoenix applications, use the module in each.
 
--> If you include use `Plug.Builder` or any other module that redefines the `call/2` function, make sure to place the `use Appsignal.Phoenix` line after that.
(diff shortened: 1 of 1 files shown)


source/elixir/installation/umbrella.html.md
```diff
@@ -27,7 +27,7 @@ By default, each nested application in an umbrella uses the main umbrella config
 The installation for Phoenix in an umbrella application are mostly the same as
 [setting up AppSignal “regular” Phoenix
 application](/elixir/integrations/phoenix.html). One thing to keep in mind is
-to `use Appsignal.Phoenix` in your application’s endpoint file. If you have
+to `use Appsignal.Phoenix` in your application’s endpoint file and add this after `Plug.Builder`. If you have
```
jeffkreeftmeijer, line 30: Most apps don’t include Plug.Builder here, so we might want to move this to a note after this paragraph:

> If you include `use Plug.Builder` or any other module that redefines the `call/2` function, make sure to place the `use Appsignal.Phoenix` line after that.

jeffkreeftmeijer (changes requested):

source/elixir/installation/umbrella.html.md
shairyar: Thanks for the tip, i will make the change.

jeffkreeftmeijer (approved):

PR #455, opened 2021-05-06
Make sleep requirement clearer for rake tasks in one-off containers

johannesluedke:
Make the sleep requirement more clear/better visible for rake tasks in one-off containers


shairyar:
Thank you @johannesluedke

shairyar (approved):

PR #475, opened 2021-08-05
Update express.html.md

arjunkomath:
Add missing dependancy in the example


tombruijn:
Hi @arjunkomath, thanks for the fix!

PR #485, opened 2021-10-05
Add plausible for analytics events

nchartiot:


source/assets/javascripts/metrics.js
```diff
@@ -0,0 +1,67 @@
+// Script for plausible.io
```
tombruijn, line 1: Would the locally sourced script be an issue if they ever update the script? Or is this a supported implementation method?

nchartiot: We would need to update the script manually. It is supported, but with this tradeoff.

Another way we can do it is to have a proxy that sources the newest script from plausible and serves it with our domain (https://plausible.io/docs/proxy/introduction), but then we need to evaluate the effort, what do you think?

tombruijn: I can tell you know that the proxy won't be missed by adblockers. I use it myself and ublock origin blocks the proxy too.

As long as they support the script being included directly it's 👍

tombruijn (approved):

source/assets/javascripts/metrics.js
nchartiot: I'm not sure how this will behave if the script is updated, but I'll keep an eye for future changes

Issue #491, opened 2021-11-02
Document configuration options for the standalone agent


unflxw:
The configuration options for the standalone agent should be documented. [The existing "Configuration" section of the docs](https://docs.appsignal.com/standalone-agent/installation.html#configuration) only describes what the configuration file should look like, but does not go into each of the available configuration options in detail. See [the configuration options list for the Ruby gem](https://docs.appsignal.com/ruby/configuration/options.html) for reference.

PR #493, opened 2021-11-04
Hostname config is now available for standalone agent

shairyar:
Customers can now use the config `hostname` when using the standalone agent.


tombruijn (approved):
Good for now. We should document this better, issue here: https://github.com/appsignal/appsignal-docs/issues/491
because now it doesn't say since which version which option is available.

luismiramirez (approved):

PR #503, opened 2021-11-19
Add requestHeaders info for Node.js

luismiramirez:
`requestHeaders` option will be available in the newest Node.js release.
Docs updated with info about it.

🚨  Please merge after releasing Node.js library with: https://github.com/appsignal/appsignal-nodejs/pull/506

### DIFF
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -785,6 +785,65 @@ options:
         config :appsignal, :config,
           request_headers: []
         ```
+    nodejs:
+      config_key: requestHeaders
+      since: 2.2.5
+      type:
+        list:
+          - string
+      default_value:
+        - accept
+        - accept-charset
+        - accept-encoding
+        - accept-language
+        - cache-control
+        - connection
+        - content-length
+        - path-info
+        - range
+        - request-method
+        - request-uri
+        - server-name
+        - server-port
+        - server-protocol
+      description: |
+        The `requestHeaders` config option contains a list of HTTP request headers which are read and stored by the AppSignal Node.js package.
+
+        This `requestHeaders` config option is an *allowlist*, which means that it will only take headers as specified by this config option. If this config option is unset it will use the AppSignal default.
+
+        Following list is the AppSignal package default.
+
+        ```js
+        client = Appsignal.new(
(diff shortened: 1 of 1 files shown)


data/config_options.yml
```diff
@@ -785,6 +785,69 @@ options:
         config :appsignal, :config,
           request_headers: []
         ```
+    nodejs:
+      config_key: requestHeaders
+      since: 2.2.5
+      since_notes:
+        - Upgrade to `2.2.5` or higher is recommended.
```
tombruijn, line 792: We can remove this. It  already says when it was added.

data/config_options.yml
```diff
@@ -785,6 +785,69 @@ options:
...
+                     "content-length",
+                     "path-info",
+                     "range",
+                     "request-method",
+                     "request-uri",
+                     "server-name",
+                     "server-port",
+                     "server-protocol"
+                   ]
+                 )
```
tombruijn, line 837: Can fix the indenting here? Just two spaces is fine.

data/config_options.yml
```diff
@@ -785,6 +785,69 @@ options:
...
+        ```
+
+        To configure AppSignal to not store any HTTP request headers on AppSignal transactions, configure the option with an empty list.
+
+        ```js
+        client = Appsignal.new(
+                   // ...Other config opts
+                   requestHeaders: []
+                 )
```
tombruijn, line 846: Same here for indenting

data/config_options.yml
```diff
@@ -785,6 +785,69 @@ options:
...
+        ```
+
+
+
```
tombruijn, line 850:
```suggestion
```

tombruijn (approved):

data/config_options.yml
luismiramirez: Ok, it's based on the other two integrations

tombruijn: That's because we had to fix some things after it was originally released. We don't want users to use the old versions of the packages.

tombruijn (approved):

unflxw (approved):

shairyar (approved):

PR #509, opened 2021-11-26
Update Elixir docs for instrument with span arg

tombruijn:
Users shouldn't set namespaces on child spans. That will either have no
effect or the span will not appear in the performance event timeline.

Even worse, in my limited testing, adding this line will cause the
sample not to show up on AppSignal.com at all.

### DIFF
--- a/source/elixir/instrumentation/instrumentation.html.md
+++ b/source/elixir/instrumentation/instrumentation.html.md
@@ -72,7 +72,7 @@ created using the `Appsignal.instrument/2` function:
 
 ```elixir
 defmodule AppsignalElixirExample do
-  def instrument do
+  def example_fn do
     Appsignal.instrument("instrument", fn ->
       :timer.sleep(500)
     end)
@@ -86,7 +86,7 @@ span. To use a different category name, use `instrument/3` instead:
 
 ```elixir
 defmodule AppsignalElixirExample do
-  def instrument do
+  def example_fn do
     Appsignal.instrument("instrument", "call.instrument", fn ->
       :timer.sleep(500)
     end)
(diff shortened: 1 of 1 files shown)


luismiramirez (approved):

source/elixir/instrumentation/instrumentation.html.md
```diff
@@ -94,15 +94,17 @@ defmodule AppsignalElixirExample do
...
 ```elixir
 defmodule AppsignalElixirExample do
   def instrument do
-    Appsignal.instrument("instrument", "call.instrument", fn span ->
-      Appsignal.Span.set_namespace(span, "custom_namespace")
-      :timer.sleep(500)
+    Appsignal.instrument("prepare_query.sql", "Build complicated SQL query", fn span ->
+      # Example of building a complex SQL query in the instrument block and performing it
+      query = "SOME complicate SQL query"
+      # The body is only set here because it's the value being calculated and instrumented in this function
```
unflxw, line 105: I don't understand what is meant by this bit here, about the body.

source/elixir/instrumentation/instrumentation.html.md
```diff
@@ -94,15 +94,17 @@ defmodule AppsignalElixirExample do
 end
 ```
 
-When passing a function that takes an argument, the `instrument/2-3` function
-calls it with the opened span to allow further customization:
+When passing a function that takes an argument, the `instrument/2-3` function calls it with the opened span to allow further customization. See the docs on hex.pm for [all available Span functions](https://hexdocs.pm/appsignal/Appsignal.Span.html).
 
 ```elixir
 defmodule AppsignalElixirExample do
   def instrument do
```
unflxw, line 101: Maybe the function should not be called `instrument` anymore?

unflxw (approved):
Much better example 👍 -- left some nitpicking, but changing it as-is would already be an improvement.

source/elixir/instrumentation/instrumentation.html.md
tombruijn: You can set the event body after the span is created, in this case an SQL query. I had trouble finding a better example, because users won't normally use `instrument` to create a root span and call `set_namespace` on it for example.

unflxw: Right, that's the event body -- I'm not sure if that's clear, that what is passed to `set_sql` is what's being referred to as "body"?

PR #517, opened 2021-12-20
Add warmup notice for people using anomaly detection as cronjob monitor

matsimitsu:
Add warmup notice for people using anomaly detection as a "has this process run" feature.

Due to the way Anomaly detection works, it's not recommended to use
warmup to detect if a process has run every 24 hours (e.g. a nightly cronjob).

This is because alerts are expired from memory if a metric hasn't been received for x amount of minutes.

Related to: https://github.com/appsignal/appsignal-server/pull/7899

## Questions:

Does the text make sense, or is there any way to make it more clear?


shairyar (approved):

shairyar:
I think it makes sense 👍🏼

PR #518, opened 2021-12-21
More GraphQL query examples

shairyar:
This PR includes an example of how to fetch the incident count of a specific app for a given date.


thijsc (changes requested):
Content looks good, but there are a lot of other commits in this pull.

PR #519, opened 2021-12-23
Fix quotes in Content Security Policy examples

Manduro:


tombruijn:
Thanks for the PR @Manduro!

PR #523, opened 2021-12-28
Add an example of creating a child span

shairyar:
This is a complete example of how to create a child span that will appear on an event timeline.


thijsc (approved):

source/nodejs/tracing/span.html.md
```diff
@@ -105,3 +112,16 @@ span.close();
 ```
 
 If the `RootSpan` is created by the core `http` integration (the most common case), the `RootSpan` is closed automatically by the integration. If any children of this `RootSpan` are left unclosed, then they are dropped from the trace.
+
+### Complete example
+
+Here is an example of how to create a child span
```
luismiramirez, line 118:
```suggestion
Here is an example of creating a child span from the current root span, adding information to it, and then closing it.
```

source/nodejs/tracing/span.html.md
```diff
@@ -105,3 +112,16 @@ span.close();
...
+
+```js
+const tracer = appsignal.tracer();
+const rootSpan = tracer.rootSpan();
+const childSpan = rootSpan.child();
+childSpan.setName(`Query.sql.model.action`);
+childSpan.setCategory("get.query");
+childSpan.close();
+```
```
luismiramirez, line 127:
```suggestion
```js
const tracer = appsignal.tracer();
const rootSpan = tracer.rootSpan();
const childSpan = rootSpan.child();

childSpan.setName(`Query.sql.model.action`);
childSpan.setCategory("get.query");
childSpan.close();
```
```

luismiramirez (approved):

PR #526, opened 2022-01-03
Add more details to the log stdout explanation

shairyar:
When setting the `log` to `stdout` the agent still creates the log file and writes to it.

### DIFF
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -498,8 +498,8 @@ options:
       - `file` (default)
         - Write all AppSignal logs to the file system.
       - `stdout` (default on [Heroku](http://heroku.com/))
-        - Print AppSignal logs in the parent process' STDOUT.
-          Useful with hosting solutions such as container systems and Heroku. The log file will still be created and agent will still write to it even if you set the log setting to STDOUT.
+        - Print AppSignal logs in the parent process' STDOUT instead of to a file.
+          Useful with hosting solutions such as container systems and Heroku.
 
       -> **Note**: At this time only the AppSignal integration supports the "stdout" log output feature. The [AppSignal agent](/appsignal/how-appsignal-operates.html), which is used by the integration, does not support "stdout" log output. It will always write to the "appsignal.log" file.
   - config_key: log_path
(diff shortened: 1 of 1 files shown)


data/config_options.yml
```diff
@@ -498,8 +498,8 @@ options:
       - `file` (default)
         - Write all AppSignal logs to the file system.
       - `stdout` (default on [Heroku](http://heroku.com/))
-        - Print AppSignal logs in the parent process' STDOUT instead of to a file.
-          Useful with hosting solutions such as container systems and Heroku.
+        - Print AppSignal logs in the parent process' STDOUT.
+          Useful with hosting solutions such as container systems and Heroku. The log file will still be created and agent will still write to it even if you set the log setting to STDOUT.
```
tombruijn, line 502:
```suggestion
          Useful with hosting solutions such as container systems and Heroku. The log file will still be created and the AppSignal agent will continue to write to it, even if you set the log setting to STDOUT.
```

data/config_options.yml
```diff
@@ -498,8 +498,8 @@ options:
...
 
       -> **Note**: At this time only the AppSignal integration supports this feature and the [AppSignal agent](/appsignal/how-appsignal-operates.html), which is used by the integration, does not yet support this.
```
tombruijn, line 504: Just reading this change again and I notice this has already been described on this line here.

```suggestion
      -> **Note**: At this time only the AppSignal integration supports the "stdout" log feature and the [AppSignal agent](/appsignal/how-appsignal-operates.html), which is used by the integration, does not yet support this.
```

shairyar: When I read that notice, it was not clear that's what it meant (agent still writes to the log file), maybe we should then change the wordings of that notice, wdyt?

tombruijn: Yes it needs to be more clear. I left another suggestion below, but feel free to improve upon it.

```suggestion
      -> **Note**: At this time only the AppSignal integration supports the "stdout" log output feature. The [AppSignal agent](/appsignal/how-appsignal-operates.html), which is used by the integration, does not support "stdout" log output. It will always write to the "appsignal.log" file.
```

thijsc (approved):

tombruijn (approved):

PR #528, opened 2022-01-07
Add sendSessionData config option docs

luismiramirez:
The `sendSessionData` config option has been added to the Node.js
integration.

**WARNING**

- Do not merge until https://github.com/appsignal/appsignal-nodejs/pull/541 is released
- Review `since` version number to match the release

### DIFF
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -988,6 +988,15 @@ options:
       default_value: false
     description: |
       Set this option to `true` to not send any session data with exception traces and performance issue samples.
+  - config_key: send_session_data
+    env_key: APPSIGNAL_SEND_SESSION_DATA
+    nodejs:
+      config_key: sendSessionData
+      since: 2.2.9
+      type: bool
+      default_value: true
+    description: |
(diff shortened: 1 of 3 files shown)


unflxw (commented):
Looks good so far, but I think it should also be mentioned wherever `skip_session_data` is mentioned in the docs, such as in ["Filter all session data"](https://docs.appsignal.com/application/session-data-filtering.html#filter-all-session-data) under "Session data filtering" and in ["Filtering options for parameters and session data"](https://docs.appsignal.com/appsignal/gdpr.html#filtering-options-for-parameters-and-session-data) under "GDPR". There may be other places where it is relevant, these are the ones I could find with a grep.

tombruijn:
What @unflxw says is correct. We should also update those pages to mention the new config option. Another page I found that it should be added to is the guide: https://docs.appsignal.com/guides/filter-data/filter-session-data.html, but that currently doesn't mention the config option for any integration yet. Only the "filter" config option to filter specific fields.

shairyar (approved):

unflxw (approved):

tombruijn (approved):

Issue #530, opened 2022-01-11
Revamp nodejs docs


shairyar:
Todos

- [x] Rename the menu called a Tracer to Custom Instrumentation. This way it follows the same naming conventions we have for Ruby and Elixir.
- [x] Merge "The Tracer object" and "Creating and using a Span" into a single docs page and call it `Instrumentation`.
- [x] Document the different methods available while creating/updating a span for example `setName`, `setCategory`, and so on. We [documented this](https://docs.appsignal.com/front-end/span.html#span-setaction-name-string) for the frontend but not for the node.js

PR #531, opened 2022-01-11
Revamp nodejs docs

shairyar:
This PR makes the following improvements to the docs for nodejs

- Rename the menu called a Tracer to Custom Instrumentation. This way it follows the same naming conventions we have for Ruby and Elixir.
-  Merge "The Tracer object" and "Creating and using a Span" into a single docs page and call it Instrumentation.
-  Document the different methods available while creating/updating a span for example setName, setCategory, and so on. We documented this for the frontend but not for the node.js

I have created a separate issue (#536 ) for instrumenting cases where out-of-the-box integration is not supported. That work will be done as part of a separate PR.

Linked with #530


source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,163 @@
...
+childSpan.setCategory("get.query");
+childSpan.setQuery(queryObj);
+```
+#### Adding metadata to a `Span`
+
+More information on this is available [here](/guides/custom-data/).
+
+#### Adding an error to a `Span`
+
+More information on this is available [here](/nodejs/tracing/exception-handling.html).
```
tombruijn, line 130:
```suggestion
For more information, see the [exception handling page](/nodejs/instrumentation/exception-handling.html).
```

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,163 @@
...
+const tracer = appsignal.tracer();
+const rootSpan = tracer.rootSpan();
+const childSpan = rootSpan.child();
+childSpan.setName("Query.sql.model.action");
+childSpan.setCategory("get.query");
+childSpan.setQuery(queryObj);
+```
+#### Adding metadata to a `Span`
+
+More information on this is available [here](/guides/custom-data/).
```
tombruijn, line 126:
```suggestion
For more information, see the [custom data guide](/guides/custom-data/).
```

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,163 @@
...
+```js
+const tracer = appsignal.tracer();
+const rootSpan = tracer.rootSpan();
+const childSpan = rootSpan.child();
+childSpan.setName("Query.sql.model.action");
+```
+
+#### Naming a `Span` category
+
+Set the category for a given `Span`. The category groups Spans together in the "Slow Events" feature, and in the "Sample breakdown".
```
tombruijn, line 102: Oh the event grouping is no longer based on the name? TIL I guess?

source/nodejs/instrumentation/scopes.html.md
```diff
@@ -0,0 +1,82 @@
...
+  // it will be the next span to be returned by `tracer.currentSpan()`!
+});
+```
+
+Once assigned to a Scope, the `Span` passed to `tracer.withSpan()` can be recalled using `tracer.currentSpan()` in a different file or lexical scope.
+
+```js
+tracer.withSpan(tracer.currentSpan(), (span) => {
+  // `span` is the same span created by `tracer.createSpan()` in the example above!
+  // http://gph.is/1KjihQe
```
tombruijn, line 40:
```suggestion
```

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,163 @@
...
+span.setName("Query.sql.model.action");
+
+// do stuff...
+
+span.close();
+```
+
+If the `RootSpan` is created by the core `http` integration (the most common case), the `RootSpan` is closed automatically by the integration. If any children of this `RootSpan` are left unclosed, then they are dropped from the trace.
+
+#### Complete example
```
tombruijn, line 148: Wow the Node.js integration is in need of an `instrument` function like the other integrations.

source/nodejs/instrumentation/instrumentation.html.md
shairyar: I actually took that explanation from [here](https://github.com/appsignal/appsignal-nodejs/blob/main/packages/nodejs/src/interfaces/span.ts#L60-L64). Let me know if that is not correct.

source/nodejs/instrumentation/instrumentation.html.md
shairyar: Yes, that will be nice to have.

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -155,6 +156,7 @@ Here is an example of creating a child span from the current root span, adding i
 const tracer = appsignal.tracer();
 const rootSpan = tracer.rootSpan();
 const childSpan = rootSpan.child();
+const queryObj = "YOUR-QUERY";
 childSpan.setName("Query.sql.model.action");
 childSpan.setCategory("get.query");
 childSpan.setQuery(queryObj);
```
luismiramirez, line 162: I've been searching through the code and I cannot find any `setQuery()` function. Does this really work?

shairyar: Just realized I typed it wrong, it is actually [setSQL](https://github.com/appsignal/appsignal-nodejs/blob/main/packages/nodejs/src/interfaces/span.ts#L107), and yes it works.

image


```
  // Span for the query
  const tracer = appsignal.tracer();
  const rootSpan = tracer.rootSpan();
  // const childSpan = tracer.createSpan({namespace: "custom-namespace"});

  const childSpan = rootSpan.child();
  const user_name = 'Shairyar';
  childSpan.setName(`Query.sql.model.action`);
  childSpan.setCategory("get.query2");
  childSpan.setSQL("select * from users where name = ?", user_name);
  childSpan.close();
```

shairyar: It has been fixed.

luismiramirez (approved):

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,166 @@
...
+
+The Tracer is responsible for tracking the currently active `Span`, and exposes functions for creating and activating new `Span`s.
+
+### Retrieving the `Tracer`
+
+```js
+const tracer = appsignal.tracer();
+```
+
+If the agent is currently inactive (you must actively set it as such by setting `active: true`), then the AppSignal client will return an instance of `NoopTracer`, which is safe to call within your code as if the agent were currently active but does not record any data.
```
unflxw, line 19:
```suggestion
If the agent is currently inactive (you must set it as such yourself, by setting `active: true`), then the AppSignal client will return a tracer object that does nothing. It is safe to call this tracer object within your code as you would if the integration was active, but it will not record any data.
```

Not sure if we should mention `NoopTracer` by name, it's an implementation detail.

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,166 @@
...
+A `Span` is the name of the object that we use to capture data about the performance of your application, any errors and any surrounding context. A `Span` can form a part of a broader **trace**, a hierarchical representation of the flow of data through your application.
+
+Each `Span` contains the following metadata:
+
+- A name
+- A start and finish timestamp
+- Tags
+- Sample data
+- The parent `Span`s ID (if any)
+- `SpanContext` identification of a `Span`
```
unflxw, line 32: This `SpanContext` is never mentioned again. Is it relevant?

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,166 @@
...
+## Creating and using a Span
+
+A `Span` is the name of the object that we use to capture data about the performance of your application, any errors and any surrounding context. A `Span` can form a part of a broader **trace**, a hierarchical representation of the flow of data through your application.
+
+Each `Span` contains the following metadata:
+
+- A name
+- A start and finish timestamp
+- Tags
+- Sample data
```
unflxw, line 30: Note that only the root span can have tags and sample data. Not sure if that should be mentioned (or not mentioned) here, or if the mention in the custom data guide linked below suffices. Below, on "Retrieving the current `RootSpan`", it is mentioned that you must retrieve the root span to add custom instrumentation -- I don't know if that refers to the tags and sample data, it's not clear to me.

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,166 @@
...
+### Creating a `Span`
+
+A `Span` can be created by calling `tracer.createSpan()`, which initializes a new `RootSpan` if there's no current `RootSpan` on the _scope_ or a `ChildSpan` of the present `RootSpan`.
+
+```js
+const tracer = appsignal.tracer();
+const span = tracer.createSpan();
+```
+
+A `RootSpan` is a type of `Span` which encapsulates the end-to-end latency for the entire request. As a `Span` is created, its start time is recorded. To give the span an end time, the `Span` must be [closed](#closing-a-span) when all the operations that the `Span` is recording are finished.
```
unflxw, line 65: Do child spans not need to be closed as well? This seems to imply closing spans is specific to the root span.

There's a section below about closing spans. Perhaps it doesn't need to be mentioned here.

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,166 @@
...
+const childSpan = tracer.createSpan(undefined, rootSpan);
+```
+
+After a `Span` is created, you can begin adding data to it using methods on the `Span` object. `ChildSpan`s and `RootSpan`s share the same interface.
+
+### Configuring a `Span`
+
+#### Naming a `Span`
+
+Set the name for a given `Span`. The Span name is used in the UI to group like requests together.
```
unflxw, line 92:
```suggestion
Set the name for a given `Span`. The Span name is used in the UI to group similar requests together.
```

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,166 @@
...
+const tracer = appsignal.tracer();
+const span = tracer.currentSpan();
+span.setName("Query.sql.model.action");
+
+// do stuff...
+
+span.close();
+```
+
+If the `RootSpan` is created by the core `http` integration (the most common case), the `RootSpan` is closed automatically by the integration. If any children of this `RootSpan` are left unclosed, then they are dropped from the trace.
```
unflxw, line 148:
```suggestion
If the `RootSpan` is created by the core `http` instrumentation (the most common case), the `RootSpan` is closed automatically by the instrumentation. If any children of this `RootSpan` are left unclosed, then they are dropped from the trace. Generally speaking, you should close the spans that you create, but not the spans that are created by the instrumentations.
```

Added some sort of guidance about which spans you should be concerned about closing.

source/nodejs/instrumentation/scopes.html.md
```diff
@@ -1,46 +1,47 @@
...
+});
 ```
 
-A `ChildSpan`, a `Span` that is the child of another `Span`, can also be created by passing an optional second argument to `tracer.createSpan()`.
+Once assigned to a Scope, the `Span` passed to `tracer.withSpan()` can be recalled using `tracer.currentSpan()` in a different file or lexical scope.
 
 ```js
-const childSpan = tracer.createSpan(undefined, span);
+tracer.withSpan(tracer.currentSpan(), (span) => {
+  // `span` is the same span created by `tracer.createSpan()` in the example above!
```
unflxw, line 39: Is this true? Looking at the implementation (`withSpan` calls `withContext` on the `ScopeManager`), it seems like [it reverts the stack to the previous span after it's done running the given function](https://github.com/appsignal/appsignal-nodejs/blob/ad40d9d8ec0dcedb38973339d75a18373eb7d3ef/packages/nodejs/src/scope.ts#L145-L152), unless there was no root span. So the span created by `tracer.createSpan()` above would not be `tracer.currentSpan()` down here.

source/nodejs/instrumentation/instrumentation.html.md
shairyar: You are correct, I will remove that as it might cause confusion.

source/nodejs/instrumentation/instrumentation.html.md
shairyar: I made an improvement to that

- The guide to adding tags now mentions they need to add tags on root span, the guide of sample data already mentions thats.
- Retrieving the root span hyperlinks to configuring span and creating child span

source/nodejs/instrumentation/instrumentation.html.md
shairyar: Child `span` needs to be closed as well and I have added that statement now.

source/nodejs/instrumentation/scopes.html.md
shairyar: This part of the docs has not been updated for a while, maybe something changed and the docs have not been updated yet. Let's finish structuring the docs and then we will take another look at it as part of separate PR.

shairyar: A new issue has been created #537

unflxw: Sounds good to me 👍 Let's assume it's true for now and fix it later if it isn't.

tombruijn (approved):
Looks good! Let's look at the content issues later.

unflxw (approved):
There's a lot of good changes here 👍 Nice work!

I'm not sure how to address this, but I think there's something that needs improvement still: we're providing the user with detailed information on how to do a bunch of things with the Tracer and Span objects... but we haven't really shown them yet why they would like to do any of that. None of this needs to be addressed in this PR, and I'm not sure what the right way to address it would be, but here's some ideas:
- Maybe the example could be moved to the top of the page. It may be necessary to add a bit more context to it, to make up for the fact that we haven't explained any of these methods yet.
- Screenshots! Spans have names, categories and whatnot, but the way these properties map to what you see in the UI isn't always intuitive. Perhaps we could use some screenshots to show these values in context, or indicate what these values map to in the UI in some other way.
- Perhaps the example could be a bit more "real-world-y"? We show you how to create a span, set properties on it and close it, with an unspecific "do stuff..." comment in between -- this could instead be wrapped in a function whose name hinted at a possible real-world use case (e.g. `function getUser(id)`) and the values used inside it could be changed accordingly (e.g. `span.setName("Query.sql.users.get")`). It doesn't need to be realistic, no need to set up actual database connections or anything -- just hinting at how it might be used in practice.

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,165 @@
...
+const tracer = appsignal.tracer();
+const rootSpan = tracer.rootSpan();
+const childSpan = rootSpan.child();
+childSpan.setName("Query.sql.model.action");
+childSpan.setCategory("get.query");
+```
+
+#### Adds sanitized SQL data as a string to a `Span`
+
+When called with a single argument, the `value` will be applied to the span as the body, which will show the sanitized query in your dashboard.
```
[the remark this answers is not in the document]
shairyar: We can improve that later, this PR just structures the nodejs documentation to follow the same pattern as ruby and elixir.

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,165 @@
...
+### Configuring a `Span`
+
+#### Naming a `Span`
+
+Set the name for a given `Span`. The Span name is used in the UI to group similar requests together.
+
+```js
+const tracer = appsignal.tracer();
+const rootSpan = tracer.rootSpan();
+const childSpan = rootSpan.child();
```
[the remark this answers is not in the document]
shairyar: A separate issue has been created for this. This PR is mostly about the structure.

source/nodejs/instrumentation/instrumentation.html.md
```diff
@@ -0,0 +1,165 @@
...
+As `Span`s represent a length of time, they must be given a finish time once all the operations that the `Span` is timing have completed. You do this by calling `span.close()`.
+
+```js
+const tracer = appsignal.tracer();
+const span = tracer.currentSpan();
+span.setName("Query.sql.model.action");
+
+// do stuff...
+
+span.close();
```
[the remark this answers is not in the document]
shairyar: Nice catch, I have updated the example.

shairyar:
> There's a lot of good changes here 👍 Nice work!
> 
> I'm not sure how to address this, but I think there's something that needs improvement still: we're providing the user with detailed information on how to do a bunch of things with the Tracer and Span objects... but we haven't really shown them yet why they would like to do any of that. None of this needs to be addressed in this PR, and I'm not sure what the right way to address it would be, but here's some ideas:
> 
> * Maybe the example could be moved to the top of the page. It may be necessary to add a bit more context to it, to make up for the fact that we haven't explained any of these methods yet.
> * Screenshots! Spans have names, categories and whatnot, but the way these properties map to what you see in the UI isn't always intuitive. Perhaps we could use some screenshots to show these values in context, or indicate what these values map to in the UI in some other way.
> * Perhaps the example could be a bit more "real-world-y"? We show you how to create a span, set properties on it and close it, with an unspecific "do stuff..." comment in between -- this could instead be wrapped in a function whose name hinted at a possible real-world use case (e.g. `function getUser(id)`) and the values used inside it could be changed accordingly (e.g. `span.setName("Query.sql.users.get")`). It doesn't need to be realistic, no need to set up actual database connections or anything -- just hinting at how it might be used in practice.

I have created a separate issue to address the above feedback. #538

tombruijn:
Merging this as is, and we'll pick up the issue that were mentioned separately.

source/nodejs/instrumentation/instrumentation.html.md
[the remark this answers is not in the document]
tombruijn: I've addressed this issue in PR #543.

PR #548, opened 2022-01-26
Document latest config changes

tombruijn:
Add notes to the config options when we added support for the
environment variable configuration. This way users can find out why the
env var doesn't work for older versions if they try to use it.

### DIFF
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -8,7 +8,7 @@ options:
     ruby:
       since: 0.3.0
       since_notes:
-        - "`0.11.6`: Support added for the system environment option."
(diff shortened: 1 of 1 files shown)


shairyar (approved):

data/config_options.yml
```diff
@@ -21,6 +21,8 @@ options:
       type: bool
       default_value:
         markdown: "`false` / detected by system"
```
unflxw, line 23: While we're at it, I think this needs changing as well: in the Node.js integration, the `active` state is not automatically set based on the presence of the push API key.

```suggestion
        markdown: "`false`"
```

unflxw (approved):

PR #549, opened 2022-01-27
Document the Elixir override config source

tombruijn:
As discussed in https://github.com/appsignal/integration-guide/pull/128
we've added the "override" config source. Document this source in the
documentation load order page. This gets linked to from the diagnose
report viewer on AppSignal.com.

Part of https://github.com/appsignal/appsignal-elixir/issues/752


shairyar (approved):

source/elixir/configuration/load-order.html.md
```diff
@@ -72,3 +73,11 @@ export APPSIGNAL_APP_NAME="AppsignalPhoenixExample"
 ```
 
 This source is listed as `env` in the [diagnose](/elixir/command-line/diagnose.html) output.
+
+##=override 5. Overrides
+
+When all the configuration is known, our integration will do a final check if the configuration doesn't contradict itself or if some defaults config options are not set.
```
luismiramirez, line 79:
```suggestion
When all the configuration is known, our integration will do a final check to find out if the configuration doesn't contradict itself or if some defaults config options are not set.
```

luismiramirez (approved):

unflxw (approved):

tombruijn:
Merging now, before the Elixir package with this config source is merged, because it only adds something and the change doesn't contradict the current way of loading the config.

PR #554, opened 2022-02-11
Add standalone agent config options

### LINKED ISSUE #546: Document standalone agent config
This section is very minimal: https://docs.appsignal.com/standalone-agent/installation.html#configuration

It should list all config options.

Like the the working dir config option, which has the difference in behavior from the integrations working dir path, that it doesn't create the directory specified automatically on start.

## To do

- [ ] Document all available standalone config options with types and behavior
    - Look out to only document the ones the standalone config listens to, it's not the same as the normal agent config https://github.com/appsignal/appsignal-agent/blob/a71f51417405898784cd5a156659a048ac1774a6/agent/src/standalone_config.rs#L48-L138
- [ ] Can we use the same `config_options.yml` file solution for this? Not required for a first version, but it may be convenient. If you're unsure how to do this, ask @tombruijn.

unflxw:
Please take a look at [the standalone config implementation](https://github.com/appsignal/appsignal-agent/blob/a71f51417405898784cd5a156659a048ac1774a6/agent/src/standalone_config.rs#L48-L138) and see if there's anything I've missed or misinterpreted. Some of the choices I made on what to document may be wrong, specifically:

- Did not document environment variables at all: in fact, made sure to not mention them at all in the standalone agent config page. This is because the only documented way to install and run the standalone agent is through the systemd integration, which hardcodes `--config /etc/appsignal-agent.conf`. My understanding is that this option disables reading information from the environment variables entirely.
- Did not document `debug_logging`, because while the value is deserialised, [it is always set to `false` in the resulting config object](https://github.com/appsignal/appsignal-agent/blob/a71f51417405898784cd5a156659a048ac1774a6/agent/src/standalone_config.rs#L93). This might be a bug, or a consequence of having deprecated that option.
- Documented `dns_servers` as a comma-separated string, because that seems to be what the agent expects. Note that the integrations expect it to be an array.
- Did not document `role` and `transmission_interval`, because they're not documented for other integrations, and frankly, I have no idea what they do. (I could kind of guess for `transmission_interval`, I suppose)

Fixes #546.

### DIFF
--- a/source/shared/_options_list.html.erb
+++ b/source/shared/_options_list.html.erb
@@ -1,6 +1,8 @@
 <% options = options_for(integration).sort_by { |o| option_value(:config_key, o, integration) || o[:named_key] } %>
 <% required_options = options.select { |o| o[:required] } %>
 <% not_required_options = options.reject { |o| o[:required] } %>
+<% show_environment = true if show_environment.nil? %>
+
 <h2>
   <a class="anchor" id="available-options"></a>
   <a href="#available-options">Available options</a>
@@ -24,9 +26,10 @@
   </li>
 </ul>
 
-<% required_options.each do |option| %>
-  <%= partial "shared/option", :locals => { :option => option, :integration => integration } %>
-<% end %>
-<% not_required_options.each do |option| %>
-  <%= partial "shared/option", :locals => { :option => option, :integration => integration } %>
+<% (required_options + not_required_options).each do |option| %>
+  <%= partial "shared/option", :locals => {
+    :option => option,
+    :integration => integration,
+    :show_environment => show_environment
+  } %>
 <% end %>
--- a/data/config_options.yml
+++ b/data/config_options.yml
@@ -118,6 +118,12 @@ options:
         Read from environment variable `NODE_ENV`, defaults to "development".
       description: |
         The environment of the app to be reported to AppSignal.
+    standalone:
+      config_key: environment
+      type: string
+      default_value: null
+      description: |
+        The environment of the app to be reported to AppSignal.
   - config_key: otp_app
     env_key: APPSIGNAL_OTP_APP
     required: true
@@ -146,6 +152,12 @@ options:
       default_value: nil
       description: |
         Name of your application as it should be displayed on AppSignal.com.
+    standalone:
+      config_key: app_name
+      type: string
+      default_value: null
+      description: |
+        Name of your application as it should be displayed on AppSignal.com.
     description: |
       -> **Note**: Changing the name or [environment](#option-env) of an existing app will create a new app on AppSignal.com.
   - config_key: push_api_key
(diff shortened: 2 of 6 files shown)


source/shared/_options_list.html.erb
```diff
@@ -1,6 +1,8 @@
 <% options = options_for(integration).sort_by { |o| option_value(:config_key, o, integration) || o[:named_key] } %>
 <% required_options = options.select { |o| o[:required] } %>
 <% not_required_options = options.reject { |o| o[:required] } %>
+<% show_environment = true if show_environment.nil? %>
```
luismiramirez, line 4:
```suggestion
<% show_environment = show_environment.nil? %>
```

luismiramirez (approved):

source/shared/_options_list.html.erb
unflxw: Wouldn't that resolve to `false` if `show_environment` was `true`?

tombruijn:
> Did not document environment variables at all: in fact, made sure to not mention them at all in the standalone agent config page. This is because the only documented way to install and run the standalone agent is through the systemd integration, which hardcodes --config /etc/appsignal-agent.conf. My understanding is that this option disables reading information from the environment variables entirely.

That's correct. The standalone agent shouldn't read from the system environment.

> Did not document debug_logging, because while the value is deserialised, [it is always set to false in the resulting config object](https://github.com/appsignal/appsignal-agent/blob/a71f51417405898784cd5a156659a048ac1774a6/agent/src/standalone_config.rs#L93). This might be a bug, or a consequence of having deprecated that option.

Since it was never documented, I think we removed the behavior in favor of `log_level` without deprecation.

> Documented dns_servers as a comma-separated string, because that seems to be what the agent expects. Note that the integrations expect it to be an array.

The integrations join the array into a comma-separated string when its set as an environment variable. We [parse the dns_servers config option](https://github.com/appsignal/appsignal-agent/blob/90e02d94d8c6993c2ddcc72ce4718050278ec443/common/src/config.rs#L724-L743) later on.

> Did not document role and transmission_interval, because they're not documented for other integrations, and frankly, I have no idea what they do. (I could kind of guess for transmission_interval, I suppose)

We added `role` for [(experimental) host grouping](https://github.com/appsignal/appsignal-agent/pull/352) (see also linked issue), but not sure if that ever got completed..
Transmission interval should not be documented.

source/shared/_options_list.html.erb
luismiramirez: Yes! Sorry, I had a brain explosion with that one.

tombruijn (approved):

source/standalone-agent/configuration-options.html.erb
```diff
@@ -0,0 +1,9 @@
+---
```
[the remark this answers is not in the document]
unflxw: Should it be replicated in the navigation as well? The navigation for the integrations (minus frontend) has the configuration be its own menu.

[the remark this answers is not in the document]
tombruijn: For now it can directly link to the subpage from the navigation item "configuration options". It's just so that we don't have to keep maintaining [this long list of redirects](https://github.com/appsignal/appsignal-docs/blob/main/netlify.toml) in the future.

PR #558, opened 2022-02-18
Fix file configuration link for each integration

shairyar:
The current link for "file configuration" takes the user to the application integration link, this PR fixes that.


tombruijn (approved):
I don't understand why that happens, but this seems like a good fix.

PR #564, opened 2022-03-01
Add a complete example code

shairyar:
This PR adds a complete example of how to send StatsD Metrics using cURL.


source/api/public-endpoint/statsd.html.md
```diff
@@ -37,4 +37,10 @@ stripe_api_duration:19|ms|#function:payment,domain:appsignal
 cpu_usage:55.1|g|#hostname:frontend1,cpu:0
 ```
 
+Here is a complete example on how to send these metrics using `cURL`.
```
tombruijn, line 40:
```suggestion
Here is a complete example on how to send these metrics using `cURL`. (Note the use of single quotes for the `--data` option and the leading `$`. Only then will line breaks work in the `--data` value.)
```

tombruijn (approved):

PR #566, opened 2022-03-14
Explain how to use the Apollo Server integration

unflxw:
The usage section for the Apollo Server integration does not explain how to use the plugin with words. Instead, it gives you a very long example of an Apollo Server setup, from which a reader with an already-working Apollo Server setup has to dig up the two lines that are relevant to them.

This change leaves the full example intact for reference, but adds a short explanation of what needs to be done, followed by those two relevant lines.

### DIFF
--- a/source/nodejs/integrations/apollo.html.md
+++ b/source/nodejs/integrations/apollo.html.md
@@ -21,6 +21,21 @@ You can then import and use the package in your app.
 
 The module includes an Apollo Server plugin for automatically instrumenting the resolvers of your application.
 
+To use the Apollo Server plugin, require it from the `@appsignal/apollo-server` package, and pass it in the `plugins` property when initialising `ApolloServer`, with an `Appsignal` instance as an argument:
+
+```js
+const { createApolloPlugin } = require("@appsignal/apollo-server");
+
+const server = new ApolloServer({
+  /* ... */
+  plugins: [createApolloPlugin(appsignal)],
+});
+```
+
+!> **NOTE:** In order for a [GraphQL query](https://www.apollographql.com/blog/the-anatomy-of-a-graphql-query-6dffa9e9e747/) to have its own entry in the Performance view of AppSignal.com, you must give it a name. For example, the named query `query AllBooks { books }` will show up in the performance view as an entry named `AllBooks`, but the unnamed query `{ books }` will be grouped together with all other unnamed queries, showing up in the performance view as a single entry named `[unknown graphql query]`.
+
+## Full example
+
 ```js
 // ENSURE APPSIGNAL IS THE FIRST THING TO BE REQUIRED/IMPORTED
 // INTO YOUR APP!
@@ -29,7 +44,7 @@ const { createApolloPlugin } = require("@appsignal/apollo-server");
 
 // You can also use one of the apollo-server integrations here,
 // e.g. `apollo-server-<integration>`. Note that you will also need to require
-// the AppSignal integration seperately.
+// the AppSignal integration for it separately.
 const { ApolloServer } = require("apollo-server");
 
 const appsignal = new Appsignal({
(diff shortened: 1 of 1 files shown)


source/nodejs/integrations/apollo.html.md
```diff
@@ -21,6 +21,21 @@ You can then import and use the package in your app.
...
+```js
+const { createApolloPlugin } = require("@appsignal/apollo-server");
+
+const server = new ApolloServer({
+  /* ... */
+  plugins: [createApolloPlugin(appsignal)],
+});
+```
+
+!> **NOTE:** You must define an [operation name](https://www.apollographql.com/blog/the-anatomy-of-a-graphql-query-6dffa9e9e747/) for your query to get an action name in the Performance view of AppSignal.com. For example, `query FetchData {}` would get the action name `FetchData` on AppSignal.com. If no operation name is set, the query will be grouped under the action name `[unknown graphql query]`.
```
tombruijn, line 35: We should link to the docs to set the action name, or add an example here how to do that.

tombruijn (approved):

source/nodejs/integrations/apollo.html.md
unflxw: I do agree that the wording is quite obtuse, I'll try and reword it tomorrow.

But there's nothing that they need to do on the application side to set the action name. GraphQL queries can have names, but these names are sent along with the query itself, not defined beforehand anywhere in the Apollo Server itself or in the AppSignal integration. What the note is trying to say is that you need to give your queries names, or else they'll be grouped together and it'll be a mess.

```graphql
# This query will show up as `[unknown graphql query]` on AppSignal,
# along with all the other unnamed queries.
{ books { title, author } }

# This query has a name, and it will show up as `AllTheBooks` on AppSignal.
query AllTheBooks { books { title, author } }
```

Screenshot 2022-03-14 at 17 57 35

You can try these examples on the `nodejs/express-apollo` test setup GraphQL playground.

tombruijn: ah right. Yes I remember this. Which also means that if their API consumers don't set a GraphQL query name, the unknown action name is reported as well. It's not something they always have control over.

unflxw: I have pushed some new wording. Please let me know if it's any better? I think it's clearer, but I'm not sure.

shairyar (approved):

source/nodejs/integrations/apollo.html.md
```diff
@@ -21,6 +21,21 @@ You can then import and use the package in your app.
...
+```js
+const { createApolloPlugin } = require("@appsignal/apollo-server");
+
+const server = new ApolloServer({
+  /* ... */
+  plugins: [createApolloPlugin(appsignal)],
+});
+```
+
+!> **NOTE:** In order for a [GraphQL query](https://www.apollographql.com/blog/the-anatomy-of-a-graphql-query-6dffa9e9e747/) to have its own entry in the Performance view of AppSignal.com, you must give it a name. For example, the named query `query AllBooks { books }` will show up in the performance view as an entry named `AllBooks`, but the unnamed query `{ books }` will be grouped together with all other unnamed queries, showing up in the performance view as a single entry named `[unknown graphql query]`.
```
tombruijn, line 35:
```suggestion
!> **NOTE:** In order for a [GraphQL query](https://www.apollographql.com/blog/the-anatomy-of-a-graphql-query-6dffa9e9e747/) to have its own entry in the Performance view of AppSignal.com, you must give it a name. For example, the named query `query AllBooks { books }` will show up in the performance view as an entry named `AllBooks`, but the unnamed query `query { books }` will be grouped together with all other unnamed queries, showing up in the performance view as a single entry named `[unknown graphql query]`.
```

tombruijn (approved):

luismiramirez (approved):

jeffkreeftmeijer (approved):

source/nodejs/integrations/apollo.html.md
unflxw: The query keyword is optional; a GraphQL statement without a prefix is assumed to be a query. 🤷

PR #568, opened 2022-04-11
Nodejs integration needs python to work

shairyar:
This PR adds a note to the docs about Python dependency.


source/nodejs/installation.html.md
```diff
@@ -6,6 +6,8 @@ Please follow the [installation guide](/guides/new-application.html) first, when
 
 ## Requirements
 
+!> Python, our dependency in Python is through [node-gyp](https://github.com/nodejs/node-gyp#on-unix), which compiles the extension for the Node.js integration. You will need a minimum version of python supported by `node-gyp`.
```
tombruijn, line 9: Can you move this to the supported operating system page instead? Make separate example for nodejs like we do for elixir and ruby where necessary.

tombruijn (approved):

PR #574, opened 2022-05-11
Add OpenTelemetry installation instructions

tombruijn:
Add instructions to add OpenTelemetry instrumentation to AppSignal. I've
added it to the installation section as it seems a better fit in the
future. This page will most likely become the new installation
instructions.

I've not linked from anywhere to this page yet as it's not released yet
and it's a beta feature that we'd like to point people to ourselves
(from a blog post or in support).

Closes https://github.com/appsignal/opentelemetry/issues/3

---

🚨 Merge after a new Node.js package has been released with OpenTelemetry support.


data/navigation.yml
```diff
@@ -485,6 +485,9 @@ sections:
               -
                 name: "Overview"
                 link: "/nodejs/installation/"
+              -
+                name: "OpenTelemetry integration <sup>Beta</sup>"
```
luismiramirez, line 489:
```suggestion
                name: "OpenTelemetry integration <sup>Experimental</sup>"
```

I'm not sure what's the procedure with these features, but as we're stating that it's experimental a couple of times on the document, it rings better to me this way.

source/nodejs/installation/opentelemetry.html.md
```diff
@@ -0,0 +1,124 @@
+---
+title: "OpenTelemetry Node.js integration <sup>Beta</sup>"
+---
+
+The AppSignal Node.js integration has experimental support for OpenTelemetry trace data. Our `@appsignal/nodejs` package ships with an OpenTelemetry processor that can transform OpenTelemetry trace data into AppSignal trace data to send to AppSignal.com. This is our first step in supporting the OpenTelemetry ecosystem in AppSignal.
```
luismiramirez, line 5: I miss a brief section explaining what OpenTelemetry is (really briefly) and pointing to their docs.

source/nodejs/installation/opentelemetry.html.md
```diff
@@ -0,0 +1,124 @@
...
+{
+   "main": "app.js",
+   "scripts": {
+     "server": "node --require ./tracing.js app.js"
+   }
+   // ...
+}
+```
+
+Now AppSignal and OpenTelemetry are integration in the app and with each other. Start the app to start receiving events from OpenTelemetry from the [libraries we support](#supported-libraries).
```
luismiramirez, line 120:
```suggestion
Now AppSignal and OpenTelemetry are integrated in the app and with each other. Start the app to start receiving events from OpenTelemetry from the [libraries we support](#supported-libraries).
```

data/navigation.yml
tombruijn: We usually tag it with Beta. We use it in other places as well. Alpha/beta/etc give people a better idea of the status than experimental. I use experimental in the actual copy to indicate it's not stable, maybe even less stable than a beta.

source/nodejs/installation/opentelemetry.html.md
```diff
@@ -0,0 +1,124 @@
...
+
+```js
+// tracing.js
+const { Appsignal, SpanProcessor } = require("@appsignal/nodejs");
+
+// Configure AppSignal
+const appsignal = new Appsignal({
+  active: true
+  // Any other AppSignal config for the app
+});
```
tombruijn, line 52: I am looking at this again and wonder if we shouldn't keep this in an `appsignal.js` file by default and import it here. If we start pushing towards an AppSignal config file we could also import it in the diagnose tool at some point to load the config by default. WDYT?

luismiramirez: I like the idea of standardizing it. Reduces confusion and if it simplifies how we get stuff for the diagnose tool.

unflxw: Sounds good, let's do it! ✅ We can start by just renaming this file `appsignal.js`. It already has the AppSignal client as its only export, which is good, because that's what the diagnose would need. 

Let's also ask users to put it at the root directory of the app, wherever the `package.json` is, instead of leaving it up to them to put it on their `src/` directory, so we can rely on that for the diagnose. That way, the instructions that are provided later, about adding `--require ./tracing.js`, are also guaranteed to be correct.

I think asking them to use `--require` to import AppSignal into their app is a good idea in general. At first, my gut reaction to this whole setup from the OpenTelemetry Getting Started guide was negative. It seemed a bit intrusive to me, given that it forces you to change your `npm run start` commands, and how you run the app in general.

But I've been thinking about it more, and it could be a lot clearer, and more importantly, a lot less error-prone than our whole deal with `// PUT THIS AT THE BEGINNING OF YOUR FILE, BEFORE ANY OTHER IMPORTS` comments in the code snippets, which going by support cases, users seem to regularly miss or ignore.

tombruijn: I've created PR #575 to discuss this change in more detail. Please leave your thoughts there :)

data/navigation.yml
```diff
@@ -485,6 +485,9 @@ sections:
...
+                link: "/nodejs/installation/opentelemetry.html"
```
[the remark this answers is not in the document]
tombruijn: You mean to have a separate page for MySQL that links back to the OpenTelemetry page?

source/nodejs/installation/opentelemetry.html.md
```diff
@@ -0,0 +1,124 @@
...
+Our integration supports the following libraries from OpenTelemetry instrumentations:
+
+- [mysql package](https://www.npmjs.com/package/mysql)
+- [mysql2 package](https://www.npmjs.com/package/mysql2)
+
+Support for more packages will follow later.
+
+## Installation
+
+First make sure to follow the [installation guide](/nodejs/installation) to add AppSignal to an application.
```
unflxw, line 22: I just noticed this now. Do we want them to do that? We're going to be asking them to require and initialise AppSignal again in `appsignal.js`. We probably should have a note that they need to change their existing installation: if they're already initialising AppSignal somewhere else (in the entrypoint of their application) they need to replace that, requiring it from `appsignal.js` instead. Otherwise I think we're going to get some not-fun-to-debug support tickets where they initialise it twice.

tombruijn: I don't want to repeat a lot of set up here. I'll explain it a bit more, but under the assumption they can follow the instructions from the docs we tell them to read.

unflxw: Yeah, you're right, repeating it all might be too much, and a hassle to maintain later.

My main worry is that if they follow the instructions in the existing document, they will end up with a `const appsignal = new Appsignal(...)` in their application's entrypoint. When they follow these new OpenTelemetry instructions, we tell them to initialise the client in `appsignal.js`. I think we should be explicit in telling them to remove it from the entrypoint.

source/nodejs/installation/opentelemetry.html.md
```diff
@@ -0,0 +1,124 @@
...
+
+At this time an app will need an [AppSignal integration for a web framework](/nodejs/installation/) like [Express](/nodejs/integrations/express.html), [Koa](/nodejs/integrations/koajs.html), etc. to create an AppSignal RootSpan. This RootSpan is used to track the OpenTelemetry events on. Our OpenTelemetry integration only support adding data to this AppSignal RootSpan at this time. This will not be necessary in the future.
```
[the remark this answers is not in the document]
tombruijn: I think without it we don't provide a lot of useful incident data, so it would be good to recommend it.

source/nodejs/installation/opentelemetry.html.md
tombruijn: I've just updated that in PR #575. So I've updated these instructions to match. Also when they have it in their `app.js` file and need to move it to the `appsignal.js` file.

data/navigation.yml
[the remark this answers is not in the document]
tombruijn: In the future (when we fully support OpenTelemetry) I don't think we should have pages that say we support MySQL if we support it only through OpenTelemetry. Assuming we've found a way to auto name events from OpenTelemetry trace data. Users can look up what libraries OpenTelemetry support and assume we do too. (It would be annoying to keep the list of supported libraries up-to-date.

> (Actually, thinking about it now, maybe it shouldn't even be linked under "Installation" yet?)

I've put it there now because it will end up there eventually, probably as part of the main install guide, and given our short term plans for it the install section seemed like a better place for it.

I'll add an MySQL nav entry for now that users can discover it this way.

source/nodejs/installation/opentelemetry.html.md
```diff
@@ -0,0 +1,133 @@
+---
+title: "OpenTelemetry Node.js integration <sup>Beta</sup>"
+---
+
+The AppSignal Node.js integration has experimental support for [OpenTelemetry](https://opentelemetry.io/) trace data. Our `@appsignal/nodejs` package ships with an OpenTelemetry processor that can transform OpenTelemetry trace data into AppSignal trace data to send to AppSignal.com. This is our first step in supporting the OpenTelemetry ecosystem in AppSignal.
+
+!> ⚠️ This integration is experimental and may break or cause undefined behavior. Use at your own risk.
```
jeffkreeftmeijer, line 7: Not too sure about this message. I could imagine users not seeing queries come in, or have weird results in AppSignal for some reason, but I don’t think we’ll break any apps because OpenTelemetry’s implementation should be solid. Could we ease on the risk a little?

```suggestion
Please let us know if you have trouble setting up the MySQL integration, if you’re seeing unexpected results, or if you have any feedback on this integration. We’re here to help!
```

source/nodejs/installation/opentelemetry.html.md
```diff
@@ -0,0 +1,133 @@
...
+Our integration supports the following libraries from OpenTelemetry instrumentations:
+
+- [mysql package](https://www.npmjs.com/package/mysql)
+- [mysql2 package](https://www.npmjs.com/package/mysql2)
+
+Support for more packages will follow later.
+
+## Installation
+
+At this time, an app will need an [AppSignal integration for a web framework](/nodejs/installation/) like [Express](/nodejs/integrations/express.html), [Koa](/nodejs/integrations/koajs.html), etc. to create an AppSignal root span. Events tracked by OpenTelemetry will be added as child spans to this root span. Our OpenTelemetry integration only supports adding events to an existing AppSignal root span at this time. This will not be necessary in the future.
```
jeffkreeftmeijer, line 28:
```suggestion
At this time, an app will need an [AppSignal integration for a web framework](/nodejs/installation/) like [Express](/nodejs/integrations/express.html) or [Koa](/nodejs/integrations/koajs.html) to create an AppSignal root span. Events tracked by OpenTelemetry will be added as child spans to this root span. Our OpenTelemetry integration only supports adding events to an existing AppSignal root span at this time. This will not be necessary in the future.
```

jeffkreeftmeijer (approved):

unflxw (approved):
[I thought I had already left an approval. My bad 😅]

tombruijn:
With everyone's approval I'm moving this to the new "waiting for release" column.

PR #575, opened 2022-05-13
Recommend the config to an appsignal.js file

tombruijn:
For Node.js and JavaScript apps, recommend moving all the config
to a `appsignal.js` file. That way we have a more predictable
structure of apps and may be able to require this file in the diagnose
command at a later stage.


source/nodejs/installation/index.html.md
```diff
@@ -47,10 +48,20 @@ const appsignal = new Appsignal({
...
+module.exports = { appsignal };
 ```
 
-!> To auto-instrument modules, the `Appsignal` module must be both **required** and **initialized** before any other package.
+You can then import and use the package in your app:
+
+```js
+// index.js
+// Place this at the top
+const { appsignal } = require("./appsignal"); // Update to the location used in the previous step
```
tombruijn, line 59: Is this clear enough, or shall we tell users to change the scripts in `package.json` like the OpenTelemetry docs do?

```json
{
  "name": "test app",
  "private": true,
  "version": "1.0.0",
  "description": "",
  "main": "app.js",
  "scripts": {
    "server": "node --require ./appsignal.js app.js"
  }
}
```

source/front-end/installation.html.md
```diff
@@ -19,27 +19,32 @@ yarn add @appsignal/javascript
...
   key: "YOUR FRONTEND API KEY",
 });
 ```
 
-It’s recommended (although not necessarily required) to use the instance of the `Appsignal` object like a singleton. One way that you can do this is by `export`ing an instance of the library from a `.js`/`.ts` file somewhere in your project.
+(It's recommended (although not necessarily required) to use the instance of the `Appsignal` object like a singleton. To do this, `export` the client from the `appsignal.js` file.)
 
-```javascript
-import Appsignal from "@appsignal/javascript";
+You can then import and use the package in your app:
```
luismiramirez, line 36:
```suggestion
You can then import and use the object in your app:
```

source/guides/configuration.html.md
```diff
@@ -123,12 +123,12 @@ config :appsignal, :config,
 
 After installation a JavaScript file with AppSignal config can be created to configure AppSignal. If no file with AppSignal configuration is present, it is most likely using [system environment variables][env] as a configuration method instead.
 
-For Node.js apps the AppSignal config or a file with AppSignal config, can be placed in any custom location, as long as it's required/imported by the app's code __before any other packages are required/imported__.
+Configuration for Appsignal for Node.js apps we recommend creating an `appsignal.js` file with the AppSignal config in the root or `src` directory of the project. Then require/import it from the app's code __before any other packages are required/imported__.
```
luismiramirez, line 126:
```suggestion
To configure Appsignal for Node.js apps we recommend creating an `appsignal.js` file with the AppSignal config in the root or `src` directory of the project. Then require/import it from the app's code __before any other packages are required/imported__.
```

source/guides/configuration.html.md
```diff
@@ -150,19 +162,31 @@ The config in the example above applies to environment Node.js is running in (`N
 
 ### Front-end JavaScript
 
-Configuration for AppSignal for front-end JavaScript can be placed in any JavaScript file or a separate `appsignal.js` file, that is then imported in the app's main JavaScript file.
+Configuration for AppSignal for front-end JavaScript we recommend creating an `appsignal.js` file with the AppSignal config in the root or `src` directory of the project. Then require/import it from the app's code __before any other packages are required/imported__.
```
luismiramirez, line 165:
```suggestion
To configure AppSignal for front-end JavaScript we recommend creating an `appsignal.js` file with the AppSignal config in the root or `src` directory of the project. Then require/import it from the app's code __before any other packages are required/imported__.
```

source/nodejs/installation/index.html.md
```diff
@@ -47,10 +48,20 @@ const appsignal = new Appsignal({
   pushApiKey: "<YOUR API KEY>" // Note: renamed from `apiKey` in version 2.2.5
 });
 
-// ...all the rest of your code goes here!
+module.exports = { appsignal };
 ```
 
-!> To auto-instrument modules, the `Appsignal` module must be both **required** and **initialized** before any other package.
+You can then import and use the package in your app:
```
luismiramirez, line 54:
```suggestion
You can then import and use the object in your app:
```

source/nodejs/installation/index.html.md
luismiramirez: I think this should be clear enough. Modifying the `package.json` file could be painful if the customer's app is a big one.

unflxw: Benefit of asking them to initialise it this way: If we ask them to use `--require`, then we make sure AppSignal gets initialised first. I have seen support cases where the user missed the "Place this at the top, before any other imports" comments. If they also need to use the AppSignal client in their code (for example, to create a span, or to use the express integration) they can require it wherever they want. Node.js modules aren't ran twice if they're required twice, so it's okay.

jeffkreeftmeijer (approved):

source/nodejs/installation/index.html.md
tombruijn: I'm merging it as-is and we can update it to use the `--require` flag in another PR. I do think I agree with Noemi that it would be a better way to do things.

tombruijn:
Also related to https://github.com/appsignal/appsignal-nodejs/issues/329

PR #578, opened 2022-05-19
Add OpenTelemetry redis and ioredis instructions

### LINKED ISSUE #577: Document OpenTelemetry redis installation instructions
Mention the need for a `dbStatementSerializer` to not send the unsanitized queries.

tombruijn:
Add documentation to instrument the redis and ioredis packages using
OpenTelemetry.

The query sanitization is a bit different between these packages, and
the listed AppSignal sanitizers don't exist (yet). If you agree we can
write default sanitizer to ship for OpenTelemetry.

The redis package sanitizer would be something like this:

```js
const sdk = new opentelemetry.NodeSDK({
  instrumentations: [
    new RedisInstrumentation({
      dbStatementSerializer: function(command, arguments) {
        const values = []
        for (let i = 0; i < arguments.length; i++) {
          values.push("?")
        }
        return `${command} ${values.join(" ")}`
      }
    }),
  ]
});
```

Which results in event bodies like `set ? ?` and `get ?`.

---

🚨 Reminder to actually implement the `dbStatementSerializer` for redis and ioredis if we agree this is what we want.

---

Closes #577


jeffkreeftmeijer (approved):

source/nodejs/integrations/opentelemetry-redis.html.md
```diff
@@ -0,0 +1,75 @@
...
+First follow the [OpenTelemetry installation instructions](/nodejs/installation/opentelemetry.html).
+
+In the `appsignal.js` file, disable the AppSignal Redis instrumentation by adding the [`instrumentRedis: false` option](/nodejs/configuration/options.html#instrumentredis) to the configuration.
+
+```js
+// appsignal.js
+// The start of the appsignal.js file, see the installation instructions
+
+const appsignal = new Appsignal({
+  instrumentRedis: false
```
tombruijn, line 18:
```suggestion
  // Other config here
  instrumentRedis: false
```

luismiramirez (approved):

source/nodejs/integrations/opentelemetry-ioredis.html.md
```diff
@@ -0,0 +1,39 @@
...
+In the `tracing.js` file, add the OpenTelemetry IORedisInstrumentation to the list of instrumentations.
+
+By default the ioredis instrumentation includes full queries in the event data sent to AppSignal, which may include Personal Identifiable Information (PII) and sensitive data. For this reason we recommend using the `IORedisDbStatementSerializer` to sanitize queries.
+
+```js
+// tracing.js
+// The start of the tracing.js file, see the installation instructions
+
+// Add the IORedisDbStatementSerializer import
+const { IORedisDbStatementSerializer } = require('@appsignal/nodejs');
```
unflxw, line 20: The `db.statement` serialiser function is imported as `RedisDBStatementSerialiser` in the example for the `node-redis` instrumentation, and as `IORedisDBStatementSerialiser` here. I believe it would be the same function in both cases, right?

source/nodejs/installation/opentelemetry.html.md
```diff
@@ -126,7 +130,12 @@ In the app's `package.json`, update the way the app is started to require the `t
 }
 ```
 
-Now AppSignal and OpenTelemetry are integrated in the app and with each other. Start the app to start receiving events from OpenTelemetry from the [libraries we support](#supported-libraries).
+Now AppSignal and OpenTelemetry are integrated in the app and with each other. Add instrumentation for any packages in the next section. When you're done with the configuration, start the app to start receiving events from OpenTelemetry from the [libraries we support](#supported-libraries).
```
unflxw, line 133: The section above in this file ("Configuring OpenTelemetry and AppSignal") still has examples specific to MySQL only. I think the installation page should be made to not be specific to any instrumentation in particular, and a separate page should be made for the MySQL-specific stuff, like there is for Redis and IORedis.

source/nodejs/integrations/opentelemetry-ioredis.html.md
```diff
@@ -0,0 +1,39 @@
+---
+title: "ioredis (OpenTelemetry)"
+---
+
+The AppSignal for Node.js integration with OpenTelemetry for ioredis, through the [`ioredis` package](https://www.npmjs.com/package/ioredis).
```
unflxw, line 5: We should describe a bit what the instrumentation provides somewhere (not necessarily in the introduction). Why would I want to integrate ioredis with AppSignal?

data/navigation.yml
```diff
@@ -551,6 +551,12 @@ sections:
               -
                 name: "node-redis (redis)"
                 link: "/nodejs/integrations/redis.html"
+              -
+                name: "node-redis (OpenTelemetry)"
+                link: "/nodejs/integrations/opentelemetry-redis.html"
+              -
+                name: "ioredis (OpenTelemetry)"
+                link: "/nodejs/integrations/opentelemetry-ioredis.html"
```
unflxw, line 559: Should we have separate pages for `node-redis` and `ioredis`? It seems to me that they could both be covered under "Redis (OpenTelemetry)", like both MySQL ones are covered under "MySQL".

Also, if we add the "(OpenTelemetry)" prefix to these ones, we should probably add it to the MySQL page as well.

unflxw (approved):
As mentioned earlier on Slack, I don't see the value provided with the Redis (as in `node-redis` 3.x, not `ioredis`) integration. Our instructions tell you to disable our existing zero-config, works-out-of-the-box Redis integration, so that you can add a lot of code to set up the OpenTelemetry Redis instrumentation, with no functional difference between the "before" and "after".

I understand that we want users to eventually move to OpenTelemetry-based instrumentation, but I don't see what the incentive would be for them to do any of this right now. I'm not opposing adding it in, but I don't see what users would gain from it.

source/nodejs/integrations/opentelemetry-ioredis.html.md
tombruijn: Yes, but I exported them separately so that we can change the implementation for either integration at any time without users having to update anything.

https://github.com/appsignal/appsignal-nodejs/pull/665/files#diff-35ec4fda19775598fbc0a5972048721feb3cbddb641a728ae562cd14b6420864R11-R12

unflxw: Awesome, that makes a lot of sense. Thank you for explaining it 👍

data/navigation.yml
tombruijn: > Also, if we add the "(OpenTelemetry)" prefix to these ones, we should probably add it to the MySQL page as well.

We'll add the mysql pages in another PR.

source/nodejs/integrations/opentelemetry-ioredis.html.md
tombruijn: Yes would be nice. I don't think it's a blocker for now seeing how little document this anyawy.

PR #593, opened 2022-05-30
Fix open graph tags

kouliavtsev:
Fixes open graph tags error by adding`<meta property=“og:type” content=“article” />` to every page.

Required open graph tags:
- title
- type 👈 added this one
- url
- image

### Debug tools
https://ahrefs.com/blog/open-graph-meta-tags/


kouliavtsev:
@aksie Is this the correct way to fix the issue? My assumption is that every page in the docs is an article. So, each `og:type` tag is an article as well.

aksie:
@kouliavtsev I am not sure, to be honest, I didn't understand from ahrefs.com what exactly was wrong with the OG tag implementation.

aksie:
Upon more detailed reading: if the TYPE indeed was missing, then article sounds like the best match for the docs. so YES! solved I Think!

aksie (approved):

PR #618, opened 2022-06-16
Add Node.js demo CLI page

tombruijn:
This page was missing, even though the CLI tool is available. I linked
from it from PR https://github.com/appsignal/appsignal-nodejs/pull/691
assuming it was there. So I've added it now.


luismiramirez (approved):

source/nodejs/command-line/demo.html.md
```diff
@@ -0,0 +1,40 @@
+---
+title: "AppSignal for Node.js: Demonstration tool"
+---
+
+The AppSignal for Node.js package ships with a command line tool used to send demonstration samples to AppSignal. Upon running it, it sends an error and performance sample to AppSignal from the user's machine.
+
+This command line tool is useful when testing AppSignal on a system and validating the local configuration. It tests if the installation of AppSignal has succeeded and if the AppSignal agent is able to run on the machine's architecture and communicate with the AppSignal servers.
+
+The same test is also run during installation using [appsignal install](/nodejs/command-line/install.html).
```
unflxw, line 9: The command is not `appsignal install` on Node, right? It would be something like `npx @appsignal/cli install`. Since that's a bit unwieldy to use in the middle of a sentence, maybe go with:

```suggestion
The same test is also run during installation using [the `install` command](/nodejs/command-line/install.html).
```

source/nodejs/command-line/demo.html.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Usage
+
+On the command line in your project run:
+
+```bash
+npx @appsignal/cli demo
+```
+
+To run it with a specific environment, Push API key or application name, see the [CLI options](#options).
```
unflxw, line 21: It would be neat to mention that the equivalent environment variables for this option (`APPSIGNAL_PUSH_API_KEY` and friends) will also be used if present.

unflxw (approved):

PR #619, opened 2022-06-16
Update index.html.md

softwaregravy:
Docs say `debug` option is deprecated and we should use `log_level` instead


shairyar:
@softwaregravy thanks for spoting that

shairyar (approved):

PR #620, opened 2022-06-17
Add example query for fetching incident with sample

shairyar:
This PR adds a GraphQL example showing how to fetch an incident with a sample.


source/api/graphql-explorer-examples.html.md
```diff
@@ -298,3 +298,85 @@ query MetricAggregationQuery($appId: String!, $start: DateTime, $end: DateTime,
...
+
+### Fetch incident with sample
+
+#### Query
+```graphql
+query IncidentQuery($appId: String!, $incidentNumber: Int!, $sampleId: String, $timestamp: String, $timerange: [DateTime]) {
+  app(id: $appId) {
+    id
+    incident(incidentNumber: $incidentNumber) {
+      ... on ExceptionIncident {
```
tombruijn, line 310: This is limited to the only exceptions. Do we also want to include error incidents? If you request an incident by id that's not an error incident it will return nothing.

source/api/graphql-explorer-examples.html.md
```diff
@@ -298,3 +298,85 @@ query MetricAggregationQuery($appId: String!, $start: DateTime, $end: DateTime,
...
+  }
+}
+```
+
+#### Query variables
+```graphql
+{
+  "appId": "YOUR-APP-ID",
+  "incidentNumber": ID-OF-INCIDENT,
+  "sampleId": "SAMPLE-ID-STRING"
```
tombruijn, line 380: This is an optional variable (according to the query variables). Can we somehow indicate this (with a comment)? Same for the timestamp and timerange variables.

source/api/graphql-explorer-examples.html.md
shairyar: Thanks for catching that, I updated the query to provide error incident and performance incident

source/api/graphql-explorer-examples.html.md
shairyar: Done

tombruijn (approved):

PR #626, opened 2022-07-13
ARM is no longer in experimental phase

shairyar:
Since ARM is no longer in the experimental phase so, this PR removes the mention of it being in an experimental


source/support/operating-systems.html.md
```diff
@@ -94,25 +92,6 @@ export APPSIGNAL_BUILD_FOR_MUSL=1
 npm/yarn install
 ```
 
-### Linux ARM build override
-
-To opt-in to the Linux ARM build manually, add the `APPSIGNAL_BUILD_FOR_LINUX_ARM` environment variable to your system environment before installing AppSignal and compiling your application. This environment variable is only needed if our auto detection does not detect the correct architecture.
```
tombruijn, line 99: This can stay, it documents the `APPSIGNAL_BUILD_FOR_LINUX_ARM` variable, which people may still find useful if our auto detection doesn't work as expected.

source/support/operating-systems.html.md
```diff
@@ -94,25 +92,6 @@ export APPSIGNAL_BUILD_FOR_MUSL=1
...
--> </ul>
-
-```sh
-# Set this environment variable
-# Note: How environment variables are set may be different for your Operating
-# System.
-export APPSIGNAL_BUILD_FOR_LINUX_ARM=1
-# Run the install command, such as: bundle install, npm install, mix compile
-```
-
```
tombruijn, line 115: This example can also stay.

shairyar: I reverted the last commit, which should undo the removal of the section.

tombruijn (approved):

PR #627, opened 2022-07-13
Use index.js instead of app.js for clarity

noccor:
Can be confusing referencing app.js and index.js (the same file) with different names. This should solve that :)


source/nodejs/installation/index.html.md
```diff
@@ -65,9 +65,9 @@ We recommend adding the `--require` flag to any script that starts your app:
   "private": true,
   "version": "1.0.0",
   "description": "",
-  "main": "app.js",
+  "main": "index.js",
   "scripts": {
-    "server": "node --require ./appsignal.js app.js"
+    "server": "node --require ./appsignal.js index.js"
```
tombruijn, line 70: We have the same example on the OpenTelemetry pages, can you update those too?
https://docs.appsignal.com/nodejs/installation/opentelemetry.html

tombruijn (approved):

PR #629, opened 2022-07-15
Correct example JS

noccor:
Tiny PR to fix an incorrect code snippet


unflxw (approved):
💯👍 Would you make this change for [the Redis instrumentation](https://github.com/appsignal/appsignal-docs/blob/main/source/nodejs/integrations/opentelemetry-redis.html.md?plain=1#L39) as well? Thank you!

PR #633, opened 2022-07-21
FAQ Doc Improvements

noccor:
Improving the FAQ page, also adds allowlisting terms: 

1. Make all subheadings questions 
2. Simplify/make some copy more readable 
3. Consistent capitalization 
4. Add allowlisting


source/faq.html.md
```diff
@@ -2,80 +2,86 @@
...
-      - [How to ignore errors in your app?](#how-to-ignore-errors-in-your-app)
-      - [How to filter data that's being send to AppSignal?](#how-to-filter-data-thats-being-send-to-appsignal)
-  - [How to add additional instrumentation to my app?](#how-to-add-additional-instrumentation-to-my-app)
-  - [How to run multiple applications on one host?](#how-to-run-multiple-applications-on-one-host)
-  - [What Operating Systems does AppSignal support?](#what-operating-systems-does-appsignal-support)
-  - [How to debug an issue with the AppSignal integration?](#how-to-debug-an-issue-with-the-appsignal-integration)
-- User account
-  - [How to enable two-factor authentication (2FA) for the AppSignal app?](#how-to-enable-two-factor-authentication-2fa-for-the-appsignal-app)
+## Contents
+- [Getting Started](#getting-started)
```
tombruijn, line 8: Okay, let me start by saying I'm not a fan of using title case in headings. For me it's actually more difficult to read. I don't know why we would want to use it here. Is there a good reason to adopt title case for _every heading_? What are the benefits?

source/faq.html.md
```diff
@@ -2,80 +2,86 @@
...
+  - [What IP Addresses does AppSignal Use?](#what-ip-addresses-does-appsignal-use)
 
-## Getting started
+## Getting Started
 
-### I need help getting started. Where do I begin?
+### I Need Help Getting Started. Where Do I Begin?
 
-If you're new to AppSignal or want to know more about to configure AppSignal features, please [read our guides](/guides/) to get set up.
+If you're new to AppSignal or want to know more about configuring AppSignal features, please [read our guides →](/guides/)
```
tombruijn, line 28: Why use the arrow symbol here instead of normal punctuation? And if we really need to use it, make it so that screen readers don't have to read it aloud. I've heard it's annoying for screen reader users.

tombruijn (changes requested):

source/faq.html.md
mmaksimovic: There is a writing standard called the Chicago Manual of Style used by publishers. We use it for the blog, and enforce it everywhere so that the style is the same.

I never pushed that to the docs, because that wasn't my focus at the time, but when Connor approached me, I agreed with using the same style everywhere, and we agreed on the industry standard. True, you may not see this within every company, but at least we would have consistency. 

For example, Ana and I spend our time making sure that every blog post follows these, but we didn't have the time to do the same for the docs.

source/faq.html.md
noccor: This is a good point (arrow key, screen readers) let me take this into consideration.

source/faq.html.md
noccor: It's not yet publicly announced, but it is in the public (AppSignal) domain that there are some formatting guidelines now for docs. This includes title case as @mmaksimovic mentioned: https://3.basecamp.com/4738529/buckets/28330015/google_documents/5147903785 - there are a lot of arguments around title case vs sentence case, both have pros and cons, but in this case it feels right to follow the example set by our blog. 

The guidelines are still being tweaked and modified but the core of what was needed in order to start applying consistency during re-writes is in there. 

A larger announcement will follow.

tombruijn: Left my reply to this on Basecamp: https://3.basecamp.com/4738529/buckets/28330015/messages/5157091142#__recording_5161130850

source/faq.html.md
roytomeij: What was your reasoning behind including the arrow @noccor?

roytomeij (approved):

source/faq.html.md
noccor: A visual hint that "this link takes you to a different page not a section within the current documentation"

roytomeij: Got it. If we want to do that everywhere, we could probably add this automatically in an accessible way using CSS or JS. I'd say we leave it out for now and fix it with code later.

noccor: Yes @roytomeij  - I've spent the past hour navigating our docs via a screen reader, we should add this in CSS.

noccor: Arrows are now removed ➡️ 🗑️ until we have a CSS solution :)

source/faq.html.md
noccor: As per my response on basecamp, for now, I'd like to keep these case changes and review once we've got more examples of documentation re-written following this style guideline and an ability. The guide section is currently being revised. This will give us an idea of casing in varying content, and we can if possible ask for 3rd party feedback.

PR #648, opened 2022-08-10
Document Ruby Garbage Collection integration

tombruijn:
We support Garbage Collection timings reported by the profiler since
Ruby gem 3.1.3 (unreleased at the time of writing).

Document the manual steps people need to take to enable AppSignal
reporting this metric.

See also [Garbage Collection profiler tracking issue][tracking issue].

Part of [tracking issue]

[tracking issue]: https://github.com/appsignal/appsignal-ruby/issues/868


tombruijn:
@noccor let met know when you want to have a quick call to run through this :)

config/navigationLinks.js
```diff
@@ -418,6 +418,10 @@ const navigationLinks = [
                 name: "Delayed::Job",
                 link: "/ruby/integrations/delayed-job.html",
               },
+              {
+                name: "Garbage Collection Profiler",
```
unflxw, line 422: Should this say "Garbage Collection", without "Profiler"? That's the name used in the document title and in the integrations index.
```suggestion
                name: "Garbage Collection",
```

source/posts/ruby/integrations/garbage-collection.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+<Notice variant="warning">
+  We do not recommend enabling the Garbage Collection profiler in production environments, at least not for extended periods of time. The additional overhead of the profiler may negatively impact the performance of your application.
+</Notice>
+
+<Notice>
+  Ruby gem version 3.1.3 or newer is required for Garbage Collection profiling.
+</Notice>
+
+When the Ruby Garbage Collection profiler is enabled AppSignal will automatically collect metrics for its Ruby VM magic dashboard. The "GC time" graph will show how long the Ruby VM took to clean up the unused memory between measurements. This graph will spike if it suddenly had to process a lot more memory, such as after a high memory usage task or request of example. By default, the GC profiler is not enabled and no metrics are reported this graph.
```
unflxw, line 19: > The "GC time" graph will show how long the Ruby VM took to clean up the unused memory between measurements.

What's the unit? If the graph says "4000", is that a time measurement, like nanoseconds or milliseconds?

Also, it would be nice to say that the measurements are once per minute. It's implied later ("after at least two minutes, after two measurements have taken place") but mentioning it along with the time unit would allow users to say "okay, so a value of 4000 in the graph means that, out of the last minute, it spent 4000 milliseconds doing garbage collection".

unflxw (approved):

source/posts/ruby/integrations/garbage-collection.mdx
tombruijn: > What's the unit? If the graph says "4000", is that a time measurement, like nanoseconds or milliseconds?

We send the data as milliseconds. I've created a new graph in https://github.com/appsignal/public_config/pull/39 that has the right value format so it's shown as a time value.

luismiramirez (approved):

noccor:
on my todo list 👀 will take a look today

source/posts/ruby/integrations/garbage-collection.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: "Garbage Collection"
+---
+
+The Ruby VM uses Garbage Collection (GC) to make sure it doesn't run out of process memory. Your application doesn't want to run out of memory and get stopped by the Operating System, resulting in the app going down.
```
noccor, line 5: Minor re-write suggestion:

_The Ruby VM (virtual machine) uses Garbage Collection (GC) to prevent it from running out of process memory. Garbage Collection stops your application from suffering memory outages due to running out of memory_

_Your application doesn't want to_ is the accidental personification of the application, so I tried to re-word that. Unsure if OS is important to include, so in my example, it's omitted.

source/posts/ruby/integrations/garbage-collection.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+This processes of cleaning up unused memory in Ruby can happen at any time. To get more insights into this we can use [Ruby's Garbage Collection profiler](https://ruby-doc.org/core/GC/Profiler.html).
```
noccor, line 7: tiny knit-pick, use "you" instead of "we" as this is not a guide

source/posts/ruby/integrations/garbage-collection.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+## Profiler
+
+<Notice variant="warning">
+  We do not recommend enabling the Garbage Collection profiler in production environments, at least not for extended periods of time. The additional overhead of the profiler may negatively impact the performance of your application.
```
noccor, line 12: Again really minor copy suggestion for brevity / de-wordy-ness:

_... at least not for extended periods. The profiler's additional overhead may negatively impact your application's performance._

As for "at least not for extended periods" it might be good to expand on that? As in, when is it OK? Or are we just saying "We really don't think you should do this unless you want a slow app"

source/posts/ruby/integrations/garbage-collection.mdx
noccor, line 19: Grammatical knit-picking, changes in bold for clarity: 

_This graph will spike if it suddenly **has** to process a lot more memory, such as after a high memory usage task or request **for** example. **The GC profiler is disabled by default, so no metrics are reported in this graph.**_

source/posts/ruby/integrations/garbage-collection.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+To enable the GC profiler, add the code shown below to your application. The profiler can be enabled at any time in the application, such as on boot or in preparation of memory intensive operations.
+
+Once the GC profiler is enabled, the "GC time" graph will start showing metrics after at least two minutes. This metric is reported by our [minutely probes](/ruby/instrumentation/minutely-probes.html) system and new measurements are made every minute.
+
+```ruby
+GC::Profiler.enable
+```
+
+To only temporarily enable the GC profiler, make sure to disable the profiler afterwards.
```
noccor, line 29: only can be omitted here for brevity.

source/posts/ruby/integrations/garbage-collection.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+```ruby
+GC::Profiler.enable
+
+# Do memory intensive operation
+
+GC::Profiler.disable
+```
+
+For the Ruby gem to report any duration over zero for the garbage collection time, the profiler needs to be enabled for more than two minutes. This is the initial time the Ruby gem needs to measure the garbage collection time. We recommend running to have it enabled for longer than the minimum two minutes, because depending on when the profiler was enabled we may not have complete measurements to compare and graph, resulting in only duration of zero seconds being reported.
```
noccor, line 39: More grammatical knit-picking:

_For the Ruby gem to report any duration over zero for the garbage collection time, the profiler **must** be enabled for more than two minutes._

could also say _"you must enable the profiler"_ for a more direct active tone. 

_We recommend running to have it enabled for longer than the minimum two minutes, because depending on when the profiler was enabled we may not have complete measurements to compare and graph, resulting in only duration of zero seconds being reported._

Is a little wordy. Recommend the following changes:

**_We recommend enabling GC for longer than the minimum of two minutes _**

Then you could alter the second half too:

_We recommend enabling GC for longer than the minimum of two minutes to ensure we have complete measurements required to compare and graph._

Is we AppSignal, or should it be you (the reader) here?

noccor (changes requested):
Looking good, I added some minor suggestions for re-wording. There's a couple of sentences that don't make sense. Let me know if my comments/suggestions need further explanation!

source/posts/ruby/integrations/garbage-collection.mdx
tombruijn: > Is we AppSignal, or should it be you (the reader) here?
We, AppSignal, can't show the graph if we don't get the data.

source/posts/ruby/integrations/garbage-collection.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: "Garbage Collection"
+---
+
+The Ruby VM (Virtual Machine) uses Garbage Collection (GC) to prevent it from running out of process memory. Garbage Collection help prevent your application from suffering memory outages due to running out of memory.
```
noccor, line 5: Tiny typo, so making a suggestion to save you having to commit a single letter

```suggestion
The Ruby VM (Virtual Machine) uses Garbage Collection (GC) to prevent it from running out of process memory. Garbage Collection helps prevent your application from suffering memory outages due to running out of memory.
```

noccor (approved):
Looks good, one tiny typo (see suggestion) and good to go!

tombruijn:
With this approved we're ready to go for the release of Ruby gem 3.1.3 🚀 
I'm not in tomorrow to release it. We can wait until next week, this change is not blocking and nothing needs to be coordinated for a timed release or anything.

We haven't decided anything yet about the [dashboard update](https://github.com/appsignal/public_config/pull/39) to make this metric visible for users that already have this dashboard, but that is also not blocking.

For more, see the [tracking issue](https://github.com/appsignal/appsignal-ruby/issues/868)

tombruijn:
I've released Ruby gem 3.1.4, but not merged this PR yet. I'm waiting for the server side updates to the dashboard for that first.

Issue #672, opened 2022-08-11
Create a pull request button broken


matsimitsu:
<img width="1030" alt="Screenshot 2022-08-11 at 10 06 30" src="https://user-images.githubusercontent.com/28408/184090521-58ac2588-4bd2-42f9-bd75-5c11e3d5adf8.png">

This button in the footer leads to a 404 on GitHub

PR #673, opened 2022-08-11
Update Create PR URL

kouliavtsev:
Fixes Create a pull request button broken #672

This link would lead into 404 land.
https://github.com/appsignal/appsignal-docs/tree/main/source/index.html.md

In Nextjs it should be this link.
https://github.com/appsignal/appsignal-docs/blob/main/source/posts/home.mdx

Implemented the link below since it's more actionable.
https://github.com/appsignal/appsignal-docs/pulls


kouliavtsev:
@matsimitsu @tombruijn 

Is https://github.com/appsignal/appsignal-docs/pulls the correct link to implement?

If not, what was the purpose of implementing the link below?
https://github.com/appsignal/appsignal-docs/tree/main/source/index.html.md

matsimitsu:
> Is https://github.com/appsignal/appsignal-docs/pulls the correct link to implement?

Yeah I think so :) 


> If not, what was the purpose of implementing the link below?

I don't know, I just noticed a broken link :P

tombruijn:
@kouliavtsev It should point to the source file of the page you're looking at. Then people can click the edit button and change things in their own fork. A path like: `https://github.com/appsignal/appsignal-docs/blob/main/source/posts/home.mdx`, where `source/posts/home.mdx` is the source file of the page.

kouliavtsev:
@tombruijn 

Makes sense now. 

I have implemented a dynamic link now.

We could change the part of URL `blob` to `edit`. This way the person visiting GitHub can start editing right away.

What I have noticed, though, is that it seems that the user can commit into the master branch. Or is it only me? Is that something we could fix by a config?

Screenshot 2022-08-11 at 12 30 06

tombruijn:
@kouliavtsev Edit also works. The only downside is that it automatically forks the repo for the signed in user, which they may not want.

They see this notice:

image

You see the direct commit on main branch option because you have write access on the repo. On another repo you see a form like this: https://github.com/github/docs/edit/main/content/index.md (Clicking this link will fork the GitHub docs repo automatically.)

image

kouliavtsev:
@tombruijn 

Okido, then let's keep as is and merge this one?

components/Footer.js
```diff
@@ -1,8 +1,7 @@
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-
 import { faGithub } from "@fortawesome/free-brands-svg-icons";
 
-export default function Footer() {
+export default function Footer({ sourceFilePath = "posts/guides/home.mdx" }) {
```
tombruijn, line 4: I don't know if there's chance of this being empty, but if there is, we can also hide this button.

tombruijn (approved):

components/Footer.js
kouliavtsev: Agree, hiding the button is better.

Issue #580, opened 2022-05-23
Document a table for rights that owners or users have


wesoudshoorn:
Create a table with rights that users/developer have.

Something in this format should work:

| Feature                   | User | Admin |
|---------------------------|------|-------|
| Invite new users          | no   | yes   |
| Create uptime monitor     | yes  | yes   |
| Create public status page | yes  | yes   |
| Monitor a new app         | yes  | yes   |
| Add billing info          | no   | yes   |

PR #686, opened 2022-08-16
Add permission table

noccor:
Relates to issue #580


noccor:
@wesoudshoorn does this cover everything? I feel like this is the most logical location for it, but could also make it it's own page.

wesoudshoorn:
I think that either @matsimitsu or @jvanbaarsen could help by searching for some permission checks in Ruby / React. This would make it easier to know exactly what people can and can not do. 

For example, I think that users can _add_ a new site, but they can't roll API keys or delete a site?

wesoudshoorn:
Converted to draft until @noccor is back

wesoudshoorn:
@noccor I removed us as reviewers because we kept getting reminders :)

PR #687, opened 2022-08-16
Implement GitHub actions

kouliavtsev:
- Copies over GitHub setup from `home` and `blog`.
- Skips test for now, due to jest is not playing along well with `jest` and `contentlyar`. Also, got some thoughts on how to do jest setup a bit differently.


tombruijn:
@kouliavtsev please try again. I've added the token.

kouliavtsev:
@tombruijn 

It seems that github is picking up `main.yml` from the `main` branch. If there a way to make it pick up the config from this branch (did I miss something)? Or should I just go ahead and merge this?

.github/main.yml
```diff
@@ -0,0 +1,61 @@
+on: [push]
```
tombruijn, line 1: @kouliavtsev I think this file should be in the `.github/workflows` sub directory. That's why it's only picking up the main branch's workflow.

kouliavtsev:
@tombruijn 

Ow, wow. Seems I had two main.yml. One not part of the workflows.yml. Thx for pointing that out.

All green now!

.github/workflows/main.yml
```diff
@@ -53,9 +53,9 @@ jobs:
...
         with:
           node-version: 14.x
-      - name: Install modules
-        run: yarn
-      - name: Build smoketest
-        run: yarn build
+      # - name: Install modules
+      #   run: yarn
+      # - name: Build smoketest
+      #   run: yarn build
```
tombruijn, line 59: If it's not needed, can you remove it?

tombruijn (commented):
There's a bit too many content changes that I don't understand. I assume they are highlighted by some part of the test suite? And that's why you fixed them? This PR does not strictly implement GitHub Actions.

.github/workflows/main.yml
kouliavtsev: Sure, let's remove comments.

source/posts/ruby/configuration/index.mdx
```diff
@@ -115,8 +115,8 @@ options](/ruby/configuration/options.html) page.
 
 ```yaml
 # config/appsignal.yml
-default: &defaults
-  # Your Push API Key, it is possible to set this dynamically using ERB. Requred
+default:
+  &defaults # Your Push API Key, it is possible to set this dynamically using ERB. Requred
```
tombruijn, line 119:
```suggestion
default: &defaults
  # Your Push API Key, it is possible to set this dynamically using ERB. Required
```
Our Ruby gem install CLI doesn't use this format, so I want to use this format in the docs to avoid confusion.

kouliavtsev: @tombruijn 

This issue is being caused by `yarn format`.

We need to update the formatting rules.

Any idea how this notation is called in yaml though (so that I could google it)?

Screenshot 2022-08-19 at 15 41 22

kouliavtsev: @tombruijn 

I think I have figured out `Node Anchors`, right?

If I look into the the prettier documentation. Then it seems that what we can do about yaml formatting is limited.

https://prettier.io/blog/2018/07/29/1.14.0.html

Should disable prettier for `appsignal-docs`, since we must preserve the formatting?

source/posts/ruby/configuration/index.mdx
```diff
@@ -116,7 +116,7 @@ options](/ruby/configuration/options.html) page.
 ```yaml
```
tombruijn, line 116:
```suggestion
<!-- prettier-ignore -->
```yaml
```

This seems to work for me to ignore any issues with code block under it.

source/posts/ruby/configuration/index.mdx
tombruijn: I think you should be able to add a `<!-- prettier-ignore -->` above the code block. See my other comment.

source/posts/ruby/configuration/index.mdx
kouliavtsev: @tombruijn 

This will render  in the view.

Screenshot 2022-08-22 at 11 08 09

Would shortening the comment this way work?

Screenshot 2022-08-22 at 11 11 36

Otherwise we could apply prettier-ignore to the whole file.

tombruijn: For me it didn't render in the YAML block because I defined it above it. But I now I see another error when I do this:

```
✘ [ERROR] [plugin @mdx-js/esbuild] Unexpected character `!` (U+0021) before name, expected a character that can start a name, such as a letter, `$`, or `_` (note: to create a comment in MDX, use `{/* text */}`)

    _mdx_bundler_entry_point-765fbd15-f626-4dcd-a433-401ba073703b.mdx:116:1:
      116 │ <!-- prettier-ignore -->
          ╵  ^

Error: Found 1 problems in 1 documents.

 └── Encountered unexpected errors while processing of 1 documents. This is possibly a bug in Contentlayer. Please open an issue.

     • "posts/ruby/configuration/index.mdx": UnexpectedMDXError: Error: Build failed with 1 error:
     _mdx_bundler_entry_point-765fbd15-f626-4dcd-a433-401ba073703b.mdx:116:1: ERROR: [plugin: @mdx-js/esbuild] Unexpected character `!` (U+0021) before name, expected a character that can start a name, such as a letter, `$`, or `_` (note: to create a comment in MDX, use `{/* text */}`)
```

I don't know if that a problem?

Either way this doesn't look like the intended behavior. Either whatever parses the MDX or prettier is wrong. It doesn't even make sense to add the "your Push API key..." comment on the line with "&defaults" on a new line.

For me your shorter suggested way still gets corrected to the "wrong" way. Maybe we should ignore this file for now in prettier? That way it doesn't check it.

```
# .prettierignore
source/posts/ruby/configuration/index.mdx
```

kouliavtsev: > Either way this doesn't look like the intended behavior. Either whatever parses the MDX or prettier is wrong. It doesn't even make sense to add the "your Push API key..." comment on the line with "&defaults" on a new line.

I think I got it fixed! 🥳 The comment issue was related  to being a `HTML` comment, which should have been a `JSX` comment.

https://prettier.io/docs/en/ignore.html#jsx

tombruijn: Ooohhh it's considered JSX! Right okay. Looks good 👍

tombruijn (approved):

PR #697, opened 2022-08-18
Curl example for reporting deployments

shairyar:
This PR adds an example of how to use the curl command to report deployments.


source/posts/application/markers/deploy-markers.mdx
```diff
@@ -90,6 +90,13 @@ To create a Deploy marker with a HTTP POST request you can use curl or some othe
 
 Read more about how to create Deploy markers with our API in our [API endpoint](/api/markers.html) documentation.
 
+### Curl example
+
```
tombruijn, line 94: I would add a little description of what this command does exactly and what they need to change to use it. But I think this is a better fit on the [markers API page](https://docs.appsignal.com/api/markers.html) linked above.

This, like the deprecated Ruby gem notify_of_deploy CLI tool that uses it, are not our recommended ways of reporting deploys. So I don't know  much we actually want to highlight this as a method. Let's continue to recommend the `revision` config option instead.

shairyar: I will move this over to the markers API page, I am adding this on Thijs's request. He was on a call with a customer who could not get the env var to work, so he wanted to show them how to do it via `curl`.

shairyar: I moved the curl example over to the page you recommended, I agree it makes more sense there since that page already explains payload.

tombruijn (approved):

source/posts/api/markers.mdx
```diff
@@ -75,6 +75,15 @@ to create by specifying the `kind` value.
 }
 ```
 
+#### Payload example with curl
+
+```bash
+curl -H "Content-Type: application/json" -XPOST https://appsignal.com/api/[app-id]/markers.json?token=[your-personal-token] -d '
+{ "marker":{ "kind": "deploy", "repository": "git@github.com:company/repository.git", "revision": "[revision]", "user": "[user]"}}'
```
tombruijn, line 82: I don't know if this linebreak is expected here?

```suggestion
curl -H "Content-Type: application/json" -XPOST https://appsignal.com/api/[app-id]/markers.json?token=[your-personal-token] \
  -d '{ "marker":{ "kind": "deploy", "repository": "git@github.com:company/repository.git", "revision": "[revision]", "user": "[user]"}}'
```
This suggestion also works

PR #698, opened 2022-08-18
Fix BETA in titles

kouliavtsev:
Re-implements https://github.com/appsignal/docs-nextjs/pull/153

That we removed during typograpy update.


kouliavtsev:
@noccor @tombruijn 

Added functinality to add `sup` tags to articles.

```mdx
---
title: "Some Article Title"
sup: "BETA" 👈
---

Article text
```

This makes it possible to render sup elements in title like this.

Screenshot 2022-08-18 at 14 40 14

Screenshot 2022-08-18 at 14 41 25

Please note: If you want the tags to be visible in the side navigation. You need to update it in the `navigationLinks.js`.

Example:

```js
{
  name: "Integration gems <sup>Deprecated</sup>",
  link: "/ruby/integrations/appsignal-gems.html",
},
```

Is this something we should add to the README.md?

tombruijn (approved):
> Is this something we should add to the README.md?

Right now I would look at existing pages that use sup-elements, but if we no longer have any in the future, I would forget it. So yes, I'd say add a small section about this to the README.

PR #773, opened 2022-10-03
Load Typography from Design System

kouliavtsev:
Should be released first: https://github.com/appsignal/design-system/pull/293

It is only possible to preview this PR locally on your computer. Don't forget to check out #293 and link to `@appsignal-tools/design-system.


kouliavtsev:
@nchartiot 

Running into some troubles with importing in `@appsignal/design-system` in `next.config.js`

```js
const { typographyConfig } = require("@appsignal/design-system");
```

I think this used to work before, yet now simply is not able to find the plugin.

Screenshot 2022-10-03 at 17 53 54

In VScode it even autocompletes the `typographyConfig` in VsCode, yet for some reason, the plugin is not being found.

Screenshot 2022-10-03 at 17 54 28

Could it be that something changed in rollup, that does not allow the plugin to work inside `next.config.js`?

nchartiot:
seems an error related to `yarn link`, did you try do run `yarn link @appsignal/design-system`? I ran it locally and it finds the config, but then the build breaks because of other reasons that I'm checking right now

kouliavtsev:
@nchartiot 

For some reason, yarn linking does not help...

In `design-system` I have tried to `yarn build` + `yarn link`.

Screenshot 2022-10-12 at 17 16 16

In `appsignal-docs` I have tried `yarn link @appsignal/design-system`

Screenshot 2022-10-12 at 17 16 32

When I go to `tailwind.config.js` and save it I get the error above.

wesoudshoorn:
@jieunization 

We are trying to consolidate the typography for the docs and the blog. For our blog, we increase the text size for larger screens. For our docs, we keep the text-size consistent.

1. Keep it as is: docs stay small, blog grows on larger screens
2. Increase text size on docs on larger screens, just like the blog
3. Decrease the text size on the blog, making it the same as the docs

I would personally vote for option 2. It would increase the font-size on larger screens from 16 to 18px, which is not a big difference but makes it slightly more readable.

jieunization:
@wesoudshoorn Yeap, I agree with you. I also vote for option 2!

kouliavtsev:
@jieunization @wesoudshoorn 

Implemented version 2, which is about to release soon. It seems that we still can keep the typography small on docs, since we are not applying `md:prose-md` class.

If we would, the typography would look like this. 

Screenshot 2022-10-28 at 14 13 01

If we would go this route, we would have to update the spacing a bit.

PR #785, opened 2022-10-14
Nodejs otel docs: fix options 

noccor:
Fix for options not displaying in Node.js documentation


noccor:
@tombruijn - could you advise here how best Anton and I proceed, and if we need to update the options.yml file to include new options in the 3.x integration?

tombruijn:
We should mark some config options in the YAML file which version of the Node.js integration they apply for I think. Something like "version: 3" and then only show for that Node.js config options page. Would that work?

kouliavtsev:
@tombruijn 

We could set the options version by getting it from the URL.

I think we got two options here on how to implement this. 

Option 1: Versions by file
Option 2: Versions by config_key

For me it's a bit hard to decide what is better. I think you are the person who could lay out the cons and pros better.

If we go for the "show options by the version in the URL route", there is another issue we have to tackle.

```
{
  url: "/elixir/configuration/options",
  integration: "elixir",
  showEnvironment: true,
},
{
  url: "/elixir/1.x/configuration/options",
  integration: "elixir",
  showEnvironment: true,
},
{
  url: "/nodejs/2.x/configuration/options",
  integration: "nodejs",
  showEnvironment: true,
},
{
  url: "/nodejs/3.x/configuration/options",
  integration: "nodejs",
  showEnvironment: true,
},
{
  url: "/ruby/configuration/options",
  integration: "ruby",
  showEnvironment: true,
},
{
  url: "/standalone-agent/configuration/options",
  integration: "standalone",
  showEnvironment: false,
},
```

/lang/1.x - prev version (elixir)
/lang - current version (elixir)
/lang/2.x - current version (nodejs)
/lang/3.x - next version

The logic for the version is a bit inconsistent. Do you have any suggestions on how to make this work?

tombruijn:
@kouliavtsev I would add a key to the YAML config option, something like `version` and filter on that when it's rendering the Node.js config options. It can be ignored for Elixir and Ruby, because it won't be set on it anyway.

The more low tech solution we can maybe do quickest is:

Render the full list of options for both versions, but for the config options that have been removed, update the description to say: "this config option was removed in version 3 favor of newConfigOptionName". That's how it's done now. We keep old config options so that users can find them and read what they need to use now instead.

noccor:
@tombruijn @kouliavtsev 

I'm merging this, and will, in a separate PR, tackle the differentiation between version-specific config options.

PR #819, opened 2022-11-01
Add TLS package info

noccor:
Adds TLS package install instructions to the syslog logging docs.


thijsc (approved):
Nice! We should look into which package managers we add instructions for. For Red-Hat style it's all yum instead of apt.

PR #828, opened 2022-11-02
Mention StatsD dependence

unflxw:
Just noticed that this implicit dependence is not actually mentioned anywhere.


source/posts/ruby/integrations/puma.mdx
```diff
@@ -14,6 +14,13 @@ The AppSignal Ruby gem automatically inserts a listener into the Puma server. No
 
 <Notice>**Note**: Puma 3.11.4 or higher required.</Notice>
 
+<Notice>
+  Since version 3.0.6 of the AppSignal gem, the StatsD server bundled with the
```
unflxw, line 18:
```suggestion
  Since version 3.0.7 of the AppSignal gem, the StatsD server bundled with the
```

luismiramirez (approved):

noccor (approved):

PR #839, opened 2022-11-08
Next / Prev Article Navigation

kouliavtsev:


kouliavtsev:
@jieunization @wesoudshoorn @noccor 

Added a simple prototype with buttons. The buttons should replace active links.

`/guides` page:

Screenshot 2022-11-08 at 13 51 48

`/guides/new-application.html` page:

Screenshot 2022-11-08 at 13 52 00

This works, yet the (next / prev) article name is being removed. Do you have any suggestions how to make this better?

kouliavtsev:
@jieunization @wesoudshoorn

Some additional thoughts.

Our end goal is to make category navigation easier for the end user.

So we are winding if it makes sense to add at top or bottom (or both) a stepper like this:

[ PREVIOUS PAGE ] [ 1 ] [ 2 ] [ 3 ] [ 4 ] [ NEXT PAGE ] 

The one from the [design-system](https://www.appsignal.design/components/steps), or something else?

jieunization:
@kouliavtsev So when you click the navigation, is it navigating inside of the left menu? I want to know what range the next & previous page works because I know there are some deeper level of pages.

kouliavtsev:
@jieunization 

Screenshot 2022-11-11 at 13 27 57

I would assume that this would stay inside the same navigation level. 

@noccor You probably know more about this than me.

kouliavtsev:
@jieunization @noccor

I was reading some other documentation. And saw this pattern.

Screenshot 2022-11-11 at 14 06 32

Would something like this work?

jieunization:
@kouliavtsev I hope we can define how we will navigate them especially at which level. 

* Are we gonna navigate them in the second level of the structure? 
* Are we gonna show the upper level after the end of the second level? 

I guess design can be different depending on how we will guide them.

kouliavtsev:
> I hope we can define how we will navigate them especially at which level.

@noccor You are probably more suitable to answer this question since you know better how the content of the documentation will be structured.

kouliavtsev:
@jieunization @noccor 

> Are we gonna navigate them in the second level of the structure?

Found an example where it seems that subsections are also part of the stepper.

Screenshot 2022-11-15 at 15 07 32

noccor:
For now, we just want to go from one step to the **Previous** or **Next** step, I think the design Anton showed is good to be worked on further @jieunization as I like that it shares the name of the next step too!

jieunization:
Okay, then I am understanding this is only for next step and previous step. 
If we want to navigate them through the left navigation, we have to think about different user cases so it will take some time to build it. 

But for now I made the design for steps. 
Screenshot 2022-11-21 at 13 22 37

Design file: https://www.figma.com/file/vZzNRZzeYx3fie8A9xKre1/stepper?node-id=0%3A1&t=Ll2ulbhFxASzD7AK-1

noccor (approved):

kouliavtsev:
Implemented design feedback:

https://www.figma.com/file/vZzNRZzeYx3fie8A9xKre1/stepper?node-id=22%3A211&t=qetIDfj1tzdB8XwZ-0

Screenshot 2022-11-23 at 12 37 28

Screenshot 2022-11-23 at 12 37 15

PR #849, opened 2022-11-10
Upgrade contentlayer

kouliavtsev:


kouliavtsev:
This PR is blocked due to a change in how the contentlayer handles the string fields. In `2.7` if you would define a string, contentlayer was ok with accepting values such as `null`, `false`, `true`, `array` etc. Actually, the string definition would work as `any`.

The affected fields `field_type`, `since_notes`, `default_value` can found in `RubyObject`, `ElixirObject`, `NodeObject`, `StandaloneObject`.


For `field_type` we could do the following:

```yml
field_type:
  list:
    - string

field_type:
  array:
    - string

# change to
field_type: array
field_type: list
```

It seems that both lists accept strings, and we output the text below on the key only:

- `<code title="Array with values of type String">Array&lt;String&gt;</code>`
- `<code title="List with values of type String">list(String)</code>`

The `since_notes` are an easy fix: `since_notes: { type: "list", of: { type: "string" } }`

The `default_value` is a place where I am a bit stuck... There are many type:

```yml
default_value: null
default_value: false
default_value: []
default_value: nil
default_value: 
  markdown: "`true` / detected by system"
```

We probably could convert the values to strings, yet this is not ideal...

```yml
default_value: "null"
default_value: "false"
default_value: "[]"
default_value: "nil"
default_value: 
  markdown: "`true` / detected by system"
```

For now, upgrading contentlayer is not crucial, yet not being able to upgrade is not ideal. I have opened an issue with a feature request. Let's wait and see what happens.

https://github.com/contentlayerdev/contentlayer/issues/331

PR #873, opened 2022-11-17
Render "On this page" server side

kouliavtsev:
Fixes: If you visit the docs you see, you briefly see no subsections by rendering the menu server side. 866#


tombruijn (approved):
I don't understand everything that's going on here, but I believe you.

PR #893, opened 2022-11-24
Fix search by implementing algolia v3

wesoudshoorn:
It seems Algolia has made a few changes which

- Disabled our scraper
- Made some new things mandatory

Fixing this in the V2 implementation did not seem to work. I started implementing V3 and started tweaking the styling.

**Done**
- The search box is pretty much done
- The CSS variables are the quickest way to get the overlay styled

**TODO**
- [ ] Test if it works
- [ ] Test if it's pulling recent data
- [x] Design the overlay through their CSS variables (you might want to export our [tailwind colors as CSS variables btw](https://gist.github.com/Merott/d2a19b32db07565e94f10d13d11a8574))

I'm in contact with them to see how our scraper should work (they made it much more complicated than plug-and-play). It is giving me results locally, but I'm still unsure which data "index" it pulls data from and if that's up to date.

@noccor it would be cool if you could test some recent pages that were added to confirm it's actually returning information.


noccor:
@wesoudshoorn @kouliavtsev  - It's not showing any results for Logging.

It is pulling in the Node 3.x docs. We also have node 2.x docs, which means you see duplicate results for each version. Can we ignore the 2.x docs for node? 

Screen Shot 2022-11-24 at 12 43 29

noccor (changes requested):
@wesoudshoorn @kouliavtsev  - It's not showing any results for Logging: https://docs.appsignal.com/logging

It is pulling in the Node 3.x docs. We also have node 2.x docs, which means you see duplicate results for each version. Can we ignore the 2.x docs for node? 

Below screenshot showing koa.js docs for 3.x and 2.x:

Screen Shot 2022-11-24 at 12 43 29

wesoudshoorn:
@kouliavtsev @noccor we should probably use this configuration: https://docsearch.algolia.com/docs/legacy/required-configuration#introduce-global-information-as-meta-tags

We can probably also use it to show relevance for specific programming languages, too.

I would like to ship a fix for the most pressing issues asap and then move forward to improving our configuration, and maybe getting rid of the Algolia logo. There are a few more things we might be able to improve our search with.

kouliavtsev:
@wesoudshoorn 

Reverted changes in this commit https://github.com/appsignal/appsignal-docs/pull/893/commits/8a59b1d2e5c9eeeb0fadd6401a010e2b2379f780 due to formatting CI not passing the check.

kouliavtsev:
@jieunization @wesoudshoorn 

Implemented the styling changes. I have tried to make the functionality look as much as AppSignal. 

Can review the changes before we ship this? 

Screenshot 2022-11-25 at 08 39 50

- The search input is a bit opinionated, and the order of the logo, placeholder, and key hints is set.
- I have tried to keep the way the input looks as close as possible to https://www.appsignal.design/components/form
- The shadow has been removed, and the x padding has been set to `0.75rem`.
- The height of the input has been updated to `36px`, which is the height that we used before.

The search modal looks like this:

Screenshot 2022-11-25 at 09 03 02

Screenshot 2022-11-25 at 09 03 49

- I have made the modal more flat by removing all the shadows.
- The search result items have box shadows with less depth to make them more subtle.
- The search input has been made more round.

noccor (approved):
Agreed @wesoudshoorn, looking good @kouliavtsev 👍  for me!

jieunization:
@kouliavtsev Thanks for building it :) 

1. We have the overlay component : https://www.appsignal.design/components/overlay
- the background color should be more darker
- I am not sure we can chance it, but probably we can have a title of the overlay box and the close button too?

2. This should be center aligned, and icon size should be smaller. 
We also have the pattern for the empty state : https://www.appsignal.design/patterns/empty-state
Screenshot 2022-11-25 at 10 16 11

3. Mobile ver: we can see there is a 'cancel' button
- 'Close' can be the better labeling (we have the cancel feature when they want to clear the keywords) 
- Can you make the 'close' as the button we have? I am not sure it's possible to change from algolia

kouliavtsev:
@jieunization 

It's a bit hard to style due to limitations by Algolia.

Screenshot 2022-11-25 at 11 21 46

1. Done
2. I have centered the logo, yet making the same [empty-state](https://www.appsignal.design/patterns/empty-state) pattern is a bit challenging. 

Because the items below `try searching for` is a list (we will have to update it) I have not centered it.

Screenshot 2022-11-25 at 11 24 43

Otherwise, it will look like this:

Screenshot 2022-11-25 at 11 27 03

3. Same as with point 2, it's a bit hard to style due to limitations by Algolia.

Let's ship this one as is, and we will try to implement the feedback at a later stage.

jieunization:
@kouliavtsev I understand it! Let's ship it the style :) 

Then can you just change the labeling to 'close'?

wesoudshoorn:
@jieunization Thanks for jumping in and providing some design feedback. Unfortunately we have some limitations due to the way this library works. To make it forward-compatible (able to upgrade to a next version) we should also be careful to customize it too much. 

@kouliavtsev great job on the styling!

PR #906, opened 2022-11-25
Add post version

kouliavtsev:


kouliavtsev:
@jieunization @noccor 

I have built the pills already on this [page](https://appsignal-docs-git-version-appsignal.vercel.app/logging). 

This looks alright. Still have to fix the spacing between the heading, pills, and notice component.

Screenshot 2022-11-25 at 12 40 39

Screenshot 2022-11-25 at 12 29 20

There is an edge case, though. When you got one badge then it looks a bit weird.

Screenshot 2022-11-25 at 12 29 43

Screenshot 2022-11-25 at 12 30 09

@noccor Do you think we always have to badges (version + beta or version + new, etc.)?

@jieunization Would you go this route when it comes to design? Or pick another design pattern?

noccor (approved):
Looks great!

wesoudshoorn:
@noccor @kouliavtsev I would like to have a bit more context to make the design for this work. As I understand it now, we have two use cases:

- Indicate a feature only works on a `VersionX+`
- Indicate feature is in beta

The tags could be a good solution for this, but I'm looking to see if we can improve the design. A simple version number might not provide enough context on this, as we can see from this screenshot below. I am thinking to combine them into a single box, with a beta/version number label and a short description. 

image

**Questions**
- Do we have just these usecases?
- Did we ever need to combine them?
- Do we need support for multiple tags?

wesoudshoorn:
@kouliavtsev @jieunization @noccor I've made some design variations

## Simple tags
- Needs more context with a hover (what does Ruby 2.1+ mean?)
- Easy to use unlimited tags

image


## Table-like tags
- Feels very structured
- Inline text
- Able to add more context to one line

image

## Message/tab mix
- Think this styling looks a bit weird, but added for contrast

image

## Sidebar tags
- We have enough space here
- Needs context with a hover state

image

jieunization:
@wesoudshoorn I am voting for 'Table-like tags' because we don't have to create any other components. Also it looks fine with one pill with the explanation!

kouliavtsev:
@wesoudshoorn 

Thx for exploring the design variations. This made me think about possibilities a bit better! 👍

Sorry for being a bit late with providing the context a bit better.

### Context

We already have built-in functionality to show `beta`, or `new`.

Screenshot 2022-11-30 at 11 29 32

You can do this like this:

```mdx
title: AppSignal Logging
sup: Beta 
```

This could also be another name, for example, `new`.

Screenshot 2022-11-30 at 11 34 58

In the articles, we got message boxes that start right below the `title`.

Screenshot 2022-11-30 at 11 36 28

When I re-opened the issues https://github.com/appsignal/appsignal-docs/pull/906#issue-1464412677, I think I made a mistake focusing on changing the `sup` being shown as a badge.


### Solution?

I like the `Message/tab mix`, not the way it looks, yet the idea behind it. 

We could do this:

Screenshot 2022-11-30 at 11 23 52

```mdx
sup: Beta
version: Ruby 14
```

Here we could copy the message component and always render it with the same text:

> This feature is only available for [version] and newer.

- The sup field works perfectly fine as does right now.
- The message boxes (version, from content) align well with each other.

Do you think this works?

@noccor @jieunization

noccor:
I prefer the table version if needed because using the Notice component makes it look like there's a warning on the page - which has a negative connotation to readers who may think they need to be alert.

PR #910, opened 2022-11-28
Add additional config info to make clear that the options do not impa…

noccor:
Adds some additional notices for our logging


unflxw (approved):

unflxw:
Thanks for taking care of this @noccor!

